PR #1, opened 2021-02-24
Update nautobot command to support new status field

dgarros:


nniehoff (approved):
LGTM

nautobot_chatops/workers/nautobot.py
```diff
@@ -222,7 +223,7 @@ def get_vlans(dispatcher, filter_type, filter_value_1):
             return False
         elif filter_type == "status":
             vlans = VLAN.objects.all()
-            choices = [(vlan.status, str(vlan.status)) for vlan in vlans]
+            choices = [(vlan.status.name, str(vlan.status.slug)) for vlan in vlans]
```
glennmatthews, line 226: nit: shouldn't need to cast `vlan.status.slug` to `str` since it's already a string.

glennmatthews (approved):

Issue #4, opened 2021-03-11
Mattermost multiple interactions failing


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

There is a bug if you run commands that require the user to select from a menu, then select something else from a second returned menu, the 2nd menu doesn't appear and the command flow simply stops.  An example is the command "get-devices", where you first select from the "menu", then select the device from the filtered list.

### Steps to Reproduce
1. In Mattermost, run command "/nautobot get-devices"
2. In the "Select a device filter" menu, select "Site" (or really any option other than "Name")

### Expected Behavior

A second menu select box should appear asking you to select a site.

### Observed Behavior

Nothing happens.  Nothing is returned and the workflow is abandoned.

On the web server, it appears the first menu is triggered via the URL ending with `/slash_command/`, but the second menu is triggered via `/interaction/` and consistently throws a 404 error.

```
[11/Mar/2021 18:17:25] "POST /api/plugins/chatops/mattermost/slash_command/ HTTP/1.1" 200 0
[11/Mar/2021 18:17:30] "POST /api/plugins/nautobot/mattermost/interaction/ HTTP/1.1" 404 7174
```

matt852:
I have identified and fixed this issue, and am working on a PR for it.

PR #6, opened 2021-03-11
Fix issue #4 with interactions failing

matt852:
Fixes issue #4 where Mattermost interactions fail when using multiple menu selections.


dgarros (approved):
Good catch, thanks

PR #7, opened 2021-03-11
Fix Webex Teams token variable name

matt852:
In the chat setup instruction docs, as well as in one warning message in a single file, the Webex Teams access token is incorrectly referred to as `webex_teams_access_token`.  The code base uses the variable name `webex_teams_token` consistently throughout the rest of the plugin.


matt852:
Fixes issue #9

dgarros (approved):
Thank you

Issue #3, opened 2021-03-11
Slack won't display more than 100 items in a dropdown list at a time


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

### Steps to Reproduce
1. Have over 100 devices in the nautobot database
2. In Slack, run the command "/nautobot get-device-facts"

### Expected Behavior

The dropdown menu where the user can select a device should include all devices in the database.

### Observed Behavior

A warning message is displayed with the following message:

Then the dropdown menu is presented, allowing me to select a device, but only 100 devices are listed.  It appears I can manually specify a device by running `/nautobot get-device-facts <device-name>`.

`More than 100 options are available. Slack limits us to only displaying 100 options at a time. `

Also note this same behavior occurs whereever a dropdown menu is used and more than 100 options are available.  For example, if you have more than 100 circuits assigned to a site, then run `/nautobot get-circuits` and filter by that site, the same issue occurs.

matt852:
I am currently working on a PR for this fix, however I wrote up the issue for tracking purposes.

matt852:
Resolved with recent v1.1.0 release

Issue #2, opened 2021-03-11
Webex Teams: Error with too large an output


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

### Steps to Reproduce
1. Have over 200 devices in one site in the Nautobot DB
2. In Webex teams, run the "get-devices" command for that category
3. For device filter, select "Site"
4. Select the site with over 200 devices in it

### Expected Behavior

For get-devices, I expected the bot to return all devices in the site.

### Observed Behavior
Webex (not nautobot) returned an error, presumably because nautobot successfully generated the list of devices, but it was too large for Webex to handle.  The error is:

`An internal error occurred: [400] Bad Request - Unable to post message to room: Message length limited to 7439 characters before encryption and 10000 characters after encryption. [Tracking ID: ROUTER_604A552B-AD26-01BB-03BA-AC12FB2803BA]`

The Tracking ID changes each time.

Also note this exact same error occurs for similar commands.  For example, if you have too many circuits at one site, and run the "get-circuits" command, and filter by "All (no filter)"

jvanderaa:
Good catch on the character length. There was something similar in the Slack dispatcher as well.

matt852:
Please assign this to me

PR #10, opened 2021-03-11
Fix slack menu limit

matt852:
This PR is to fix issue #3.

Essentially this was the easiest way I could think of to add support for menu choices where there were more than 100 options without doing additional refactoring of how the dispatchers are written.

The default behavior in Slack now is if more than 100 choices are present, the last choice is replaced with the option `More...` for the user.  If the user selects `More...`, then the menu is redisplayed to the user with the next 100 choices.  If there are no more choices, `More...` doesn't appear.

### DIFF
--- a/nautobot_chatops/dispatchers/slack.py
+++ b/nautobot_chatops/dispatchers/slack.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import os
 import time
 
 from django.conf import settings
@@ -37,7 +38,7 @@ class SlackDispatcher(Dispatcher):
         """Init a SlackDispatcher."""
         super().__init__(*args, **kwargs)
         self.slack_client = WebClient(token=settings.PLUGINS_CONFIG["nautobot_chatops"]["slack_api_token"])
-        self.slack_menu_limit = 100
+        self.slack_menu_limit = int(os.getenv("SLACK_MENU_LIMIT", 100))
 
     @classmethod
     @BACKEND_ACTION_LOOKUP.time()
--- a/nautobot_chatops/dispatchers/slack.py
+++ b/nautobot_chatops/dispatchers/slack.py
@@ -36,6 +36,7 @@ class SlackDispatcher(Dispatcher):
         """Init a SlackDispatcher."""
         super().__init__(*args, **kwargs)
         self.slack_client = WebClient(token=settings.PLUGINS_CONFIG["nautobot_chatops"]["slack_api_token"])
+        self.slack_menu_limit = 100
 
     @classmethod
     @BACKEND_ACTION_LOOKUP.time()
@@ -272,20 +273,32 @@ class SlackDispatcher(Dispatcher):
         # In Slack, a textentry element can ONLY be sent in a modal dialog
         return self.send_blocks(blocks, callback_id=action_id, ephemeral=True, modal=True, title=title)
 
-    def prompt_from_menu(self, action_id, help_text, choices, default=(None, None), confirm=False):
+    def prompt_from_menu(self, action_id, help_text, choices, default=(None, None), confirm=False, tracker=0):
         """Prompt the user for a selection from a menu.
 
         Args:
           action_id (str): Identifier string to attach to the "submit" action.
           help_text (str): Markdown string to display as help text.
           choices (list): List of (display, value) tuples
-          default (tuple): Default (display, valud) to pre-select.
+          default (tuple): Default (display, value) to pre-select.
           confirm (bool): If True, prompt the user to confirm their selection (if the platform supports this)
+          tracker (int): If set, starts displaying choices at index location from all choices,
+                         as Slack only displays 100 options at a time
         """
-        # TODO: Slack limits an option list to no more than 100 items. How can we work around this?
-        if len(choices) > 100:
-            self.send_warning("More than 100 options are available. Slack limits us to the first 100.")
-            choices = choices[:100]
+        choice_length = len(choices)
+        if choice_length > self.slack_menu_limit:
+            self.send_warning(
+                f"More than {self.slack_menu_limit} options are available. Slack limits us to only displaying {self.slack_menu_limit} options at a time."
+            )
+            new_tracker = tracker + self.slack_menu_limit - 1
+            if tracker == 0:
+                choices = choices[: self.slack_menu_limit - 1]  # 1 is to leave space for 'next' insert
+            else:
+                choices = choices[tracker:new_tracker]
+            if choice_length > new_tracker:
+                # Only insert a 'next' tracker if we still have more choices left to see
+                choices.append(("Next...", f"menu_track...{new_tracker}"))
+
         menu = self.select_element(action_id, choices, default=default, confirm=confirm)
         cancel_button = {
             "type": "button",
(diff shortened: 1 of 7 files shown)


nautobot_chatops/dispatchers/slack.py
```diff
@@ -272,20 +273,32 @@ def prompt_for_text(self, action_id, help_text, label, title="Your attention ple
...
+        choice_length = len(choices)
+        if choice_length > self.slack_menu_limit:
+            self.send_warning(
+                f"More than {self.slack_menu_limit} options are available. Slack limits us to only displaying {self.slack_menu_limit} options at a time."
+            )
+            new_tracker = tracker + self.slack_menu_limit - 1
+            if tracker == 0:
+                choices = choices[: self.slack_menu_limit - 1]  # 1 is to leave space for 'next' insert
+            else:
+                choices = choices[tracker:new_tracker]
```
glennmatthews, line 297: Instead of `tracker`, consider something like `offset` perhaps? That would match the terminology used, for example, for pagination in the REST API.

nautobot_chatops/workers/nautobot.py
```diff
@@ -195,7 +211,7 @@ def get_vlans(dispatcher, filter_type, filter_value_1):
     if not filter_type:
         prompt_for_vlan_filter_type("nautobot get-vlans", "select a vlan filter", dispatcher)
         return False
-    if not filter_value_1:
+    if not filter_value_1 or is_menu_track_item(filter_value_1):
```
glennmatthews, line 214: A general nitpick - I find logic of the form `if not x or y` to be a bit non-obvious. Would you be able to change these throughout to `if y or not x` (such as `if is_menu_track_item(filter_value_1) or not filter_value_1`) and still have things work as intended?

glennmatthews (commented):
Looks like a clever fix to the issue! It seems like this general design should be possible to follow to address issues like #2 as well. Would it be feasible to add some unit test automation to cover the new `prompt_from_menu` behavior?

nautobot_chatops/dispatchers/slack.py
matt852: Sure thing.  I've renamed all occurrences of `tracker` to `offset`.  Same thing for relevant function names.

nautobot_chatops/workers/nautobot.py
matt852: This has been reworked and is much cleaner now in commit Rework "if not x or y" logic 6d8faf3.  Good idea thank you.

matt852:
I was getting this pylint error:

`nautobot_chatops/workers/nautobot.py:1:0: C0302: Too many lines in module (1024/1000) (too-many-lines)`

So I moved some basic/static functions into a new file `helper_functions.py`.

I also had to rework the passing of function variables to `prompt_for_menu`, as I was hitting this pylint error:

`R0913: Too many arguments (7/6) (too-many-arguments)`

matt852:
@glennmatthews Regarding the comment on adding unittests, I started writing them ideally the unittests for dispatchers needs to be restructured.  I have this written up in a separate PR, and will include unittests for `prompt_from_menu` in that.

glennmatthews:
I'd vote for just disabling the `too-many-arguments` warning on the `prompt_for_menu` function, rather than making a backwards-incompatible change to the function parameters:

```python
def prompt_from_menu(  # pylint: disable=too-many-arguments
    self, action_id, help_text, choices, default=(None, None), confirm=False, offset=0
):
```

matt852:
> I'd vote for just disabling the `too-many-arguments` warning on the `prompt_for_menu` function, rather than making a backwards-incompatible change to the function parameters:
> 
> ```python
> def prompt_from_menu(  # pylint: disable=too-many-arguments
>     self, action_id, help_text, choices, default=(None, None), confirm=False, offset=0
> ):
> ```

Those changes were reverted back.

nautobot_chatops/workers/helper_functions.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def menu_item_check(item):
+    """Return True if value starts with 'menu_offset' for dealing with Slack menu display limit."""
+    if item:
+        try:
+            return item.startswith("menu_offset-")
+        except AttributeError:
+            return False
+    return True
```
glennmatthews, line 20: Logic seems wrong here to me:

```python
>>> menu_item_check("menu_offset-1")
True
>>> menu_item_check("hello")
False
>>> menu_item_check("")
True
>>> menu_item_check(None)
True
```

Shouldn't we be returning `False` for an empty string or `None`?

nautobot_chatops/workers/nautobot.py
```diff
@@ -422,16 +377,24 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
...
+                choices,
+                offset=menu_offset_value(filter_type),
             )
         else:
-            dispatcher.prompt_from_menu(f"nautobot get-interface-connections {filter_type}", "Select a site", choices)
+            dispatcher.prompt_from_menu(
+                f"nautobot get-interface-connections {filter_type}",
+                "Select a site",
+                choices,
+                offset=menu_offset_value(filter_type),
```
glennmatthews, line 392: This seems wrong to me - if you get to the `else` case, here, you've already established that `filter_type == "device"`. Should any of the above block be using `filter_value_1` instead of `filter_type`?

nautobot_chatops/workers/helper_functions.py
matt852: > if not filter_value_1 or is_menu_track_item(filter_value_1):
>
> A general nitpick - I find logic of the form if not x or y to be a bit non-obvious. Would you be able to change these throughout to if y or not x (such as if is_menu_track_item(filter_value_1) or not filter_value_1) and still have things work as intended?

This was created specifically to address the above comment, and condense the logic into a single function.

nautobot_chatops/workers/nautobot.py
matt852: Great catch, thank you.  I fixed this in the latest commit.
Also while troubleshooting this, I came across 2 more bugs.  I'll put them in a separate issue/PR.

jvanderaa:
@matt852 are we all set on this one?

matt852:
I believe @glennmatthews still wanted to review it once more.

glennmatthews (approved):
Looks good. Thanks for this, and sorry for the delays!

nautobot_chatops/dispatchers/slack.py
```diff
@@ -36,6 +37,7 @@ def __init__(self, *args, **kwargs):
         """Init a SlackDispatcher."""
         super().__init__(*args, **kwargs)
         self.slack_client = WebClient(token=settings.PLUGINS_CONFIG["nautobot_chatops"]["slack_api_token"])
+        self.slack_menu_limit = 100
```
jvanderaa, line 40: Let's move this is to a configurable setting.

```suggestion
        self.slack_menu_limit = int(os.getenv("SLACK_MENU_LIMIT", 100))
```

jvanderaa (changes requested):
One small change.

nautobot_chatops/dispatchers/slack.py
matt852: Updated

jvanderaa (approved):

matt852:
The TravisCI failure looks like an anomaly, and should work if reran.

jvanderaa (approved):

Issue #11, opened 2021-03-11
get-rack command output is unordered


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

The site menu selection displayed for filtering, when runningr the command `get-rack`, is unordered, and all filter choices appear randomly.  This should be fixed to return all results in an ordered manner, similar to all other commands with selection menus.

### Steps to Reproduce
1. Create multiple sites in Nautobot
2. Create multiple racks in multiple sites in Nautobot
3. Run command `/nautobot get-rack`

### Expected Behavior

All sites returned in the "Select a Site" filter selection box should be ordered.

### Observed Behavior

All sites returned in the "Select a Site" filter selection box appear randomized in no particular order.

PR #12, opened 2021-03-11
Fix site ordering for get-rack command filter

matt852:
This is a fix for issue #11 .

In addition, I've also rearranged a few of the static filters displayed to users to be consistent in their ordering, and to move any options for `All (no filter)` to the bottom.  Currently some filters have `All (no filter)` at the top, and others at the bottom.  This keeps it consistent across the board as well.


dgarros (approved):
Looks good to me, thank you for the fix

Issue #20, opened 2021-03-16
Error filtering interface connections by site


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

When running /nautobot get-interface-connections and filtering by "Site", the following error occurs:
`:warning: I don't know how to filter by site :warning:`

### Steps to Reproduce
1. Run /nautobot get-interface-connections
2. Select "Site" to filter by

### Expected Behavior

A list of sites to appear to filter by

### Observed Behavior
The following error message occurs and the workflow is terminated:
`:warning: I don't know how to filter by site :warning:`

matt852:
Please assign this to me.  I have a PR for this fix.

matt852:
Merged

PR #21, opened 2021-03-16
#20 - Fix filtering interface connections by Site

matt852:
Fix for issue #20 when filtering device connections by Site.

Current version.  Both calls should return True.

```
>>> filter_type = "device"
>>> ("device" or "site") in filter_type
True
>>> filter_type = "site"
>>> ("device" or "site") in filter_type
False
```

Updated version in this PR:
```
>>> filter_type = "device"
>>> filter_type in ["device", "site"]
True
>>> filter_type = "site"
>>> filter_type in ["device", "site"]
True
>>>
```


glennmatthews (approved):
Thanks for the fix!

jvanderaa (approved):

Issue #14, opened 2021-03-12
Update Installation Docs


jvanderaa:
### Environment
* Python version:  3.8.6
* Nautobot version: 1.0.0b2
* nautobot-plugin-chatops version: 1.0.0

### Steps to Reproduce
1. Run through the installation docs
2.
3.

### Expected Behavior

Installation successful

### Observed Behavior

Slight modification of the docs were needed

PR #29, opened 2021-04-01
Update msteams docs zip

matt852:
This has 2 parts:

I updated the docs for installing the ChatOps bot with MS Teams to clarify some steps, and expand on some others.  Additionally, some of the windows/tabs/menu names changed in Azure since originally written, so they too have been updated.

Secondly, I regenerated the Nautobot_ms_teams.zip file to include the latest manifest.json version (was using 1.7, now using 1.9), removed the command "kentik", and replaced the command "netbox" with "nautobot".  The two .png files remain unchanged.  I also updated the command scopes from just "personal" and "team" to also include "groupchat".

This can be tagged onto issue #14, or a separate MS Teams dedicated issue can be created if preferred.


dgarros (approved):
Thanks for this PR @matt852, I noticed the `Nautobot_ms_teams.zip` was outdated recently and I wanted to open an issue for it.

jvanderaa (approved):

PR #32, opened 2021-04-13
#2 - Fix Webex Teams error on too large output

matt852:
This fixes issue #2.  Currently WebEx Teams throws a literal error if the output is too large.  However unlike Slack, where the output is limited by the number of data rows, WebEx Teams limits the output to specifically 7439 characters or less.

Therefore, I've added in logic to add as many rows as possible into each snippet posted, and only draw the headers for the `texttables` package on the first snippet.  This allows a user with multiple outputs to be able to copy/paste all of it at once, without dealing with multiple headers.


nautobot_chatops/dispatchers/webex_teams.py
```diff
@@ -112,3 +117,44 @@ def user_mention(self):
...
+            table.set_cols_dtype(["t" for item in header])
+
+            while True:
+                try:
+                    table.add_rows([rows[row_counter]], header=False)
+                except IndexError:
+                    # We're at the end of the rows being output.  Break loop
+                    break
+                table_length = len(table.draw())
+                # Length of table has exceeded maximum allowable characters by Webex.  Break loop
```
glennmatthews, line 147: Would it be cleaner/simpler to simply split the table when reaching a `table_length` of `webex_msg_char_limit - 120`, rather than exceeding the limit and having to back up and regenerate the table with one less row?

matt852: I didn't think of this.  Updating the code as you're right, it's much simpler.

nautobot_chatops/tests/test_dispatchers.py
```diff
@@ -154,6 +155,24 @@ def test_get_prompt_from_menu_choices(self):
...
+        for i in range(0, 300):
+            rows.append((f"Switch0{i}", "Active", "", "test01", "", "role01", "3560", "1.2.3.4"))
+            if i >= 200:
+                expected_arg0 += f"Switch0{i}   Active            test01          role01   3560   1.2.3.4   \n"
+        expected_arg0 += "\n```"
+
+        self.dispatcher.send_large_table(header, rows)
+
+        self.assertTrue(mock_send_markdown.called)
+        self.assertEqual(mock_send_markdown.call_args[0][0], expected_arg0)
```
glennmatthews, line 174: This test is a little hard for me to follow. I assume the intent is to verify that `send_markdown` is called multiple times, each time with an appropriate amount of content?

If so, maybe something like the following might be clearer or more thorough?

```python
header = ["Name", "Status", "Tenant", "Site", "Rack", "Role", "Type", "IP Address"]
rows = []
for i in range(0, 300):
    rows.append((f"Switch0{i}", "Active", "", "test01", "", "role01", "3560", "1.2.3.4"))

self.dispatcher.send_large_table(header, rows)

mock_send_markdown.assert_called()

total_message_length = 0
# Make sure no individual call exceeded the Webex message length limit
for call_args in mock_send_markdown.call_args_list:
    self.assertLessEqual(len(call_args[0]), self.dispatcher.webex_msg_char_limit
    total_message_length += len(call_args[0])

# Make sure the total message length was in excess of the Webex limit
self.assertGreater(total_message_length, self.dispatcher.webex_msg_char_limit)

# Make sure content was appropriately divided amongst the separate messages:
# Header should only be in the first call, not in subsequent ones
self.assertIn("Name ", mock_send_markdown.call_args_list[-1][0])
for call_args in mock_send_markdown.call_args_list[:-1]:
    self.assertNotIn("Name ", call_args[0]
# Make sure the first row appears in the first send, and the last row appears in the last send
self.assertIn("Switch00 ", mock_send_markdown.call_args_list[-1][0])
self.assertIn("Switch0299 ", mock_send_markdown.call_args_list[0][0])
```

glennmatthews (approved):
Nice fix!

nautobot_chatops/tests/test_dispatchers.py
matt852: With the data used to generate the table rows, exactly 100 rows is generated before hitting the character limit, 7408 out of 7439 total characters.
The test was generated in that the last output sent from Nautobot to the MS Teams client should _only_ contain devices Switch0200-0299, with any extra ` or whitespaces added as needed.

I'll play around with your ideas and see what works.

matt852: @glennmatthews I updated the unittests with some of your ideas if you want to re-review.

glennmatthews: Looks great, thanks!

jvanderaa (approved):

PR #34, opened 2021-04-15
Documentation update Fixes #27

### LINKED ISSUE #27: Doc Update: Slack need to invite the bot into the channel
### Environment
* Python version: n/a
* Nautobot version:  1.0.0b2
* nautobot-plugin-chatops version: 1.0.0

### Steps to Reproduce
1. Followed setup instructions
2. Troubleshot error with the bot
3. Found that the bot needed to be invited in

### Expected Behavior

Would expect that this would be in the documentation

### Observed Behavior

The bot was not fully usable with the docs as is

dgjustice:
- Moved all the install text to the chat_setup document
- Formatting changes to make mdlint happy
- Add a couple images for eye candy
- Add a note about inviting the chatbot in Slack

Fixes #27


docs/chat_setup.md
```diff
@@ -1,5 +1,37 @@
...
+```
+
+Nautobot supports `Slack`, `MS Teams` and `Webex Teams` as backends but by default all of them are disabled. You need to explicitly enable the chat platform(s) that you want to use in the `PLUGINS_CONFIG` with one or more of `enable_slack`, `enable_ms_teams` or `enable_webex_teams`.
+
+The plugin behavior can be controlled with the following list of general settings:
+
+| Configuration Setting        | Description | Mandatory? | Default |
+| ---------------------------- | ----------- | ---------- | ------- |
+| `delete_input_on_submission` | After prompting the user for additional inputs, delete the input prompt from the chat history | No | `False` |
+
```
jvanderaa, line 34: I'd like to see this left where it was on the main page, keeping the plugin documentation consistent on the install instructions personally. I'm also ok if this moves to here.

This is something else that I'm noticing as part of this PR is that `Mattermost` is missing from the list of backends, which is now supported.

jvanderaa (changes requested):
At a minimum the one change to include `Mattermost`, if not the other about leaving installation.

docs/chat_setup.md
jvanderaa: I think we have enough present. My original comment was to keep install locally in the main readme. My mind is changed on that front!

dgjustice: Can we have the author of the Mattermost change update the documentation appropriately?  I have never used the platform and know nothing of the setup.

jvanderaa: I'll double check on the setup if it is there. But the front page on line 27 of the readme should have that update.

dgjustice: Oh, I see what you are saying now.

jvanderaa (approved):

jvanderaa (changes requested):
Let's remove the install instructions from primary readme since it is moved to the install items.

jvanderaa (approved):

PR #37, opened 2021-04-20
Updates for Nautobot B4

jvanderaa:
Updates dev environment to B4

### DIFF
--- a/development/nautobot_config.py
+++ b/development/nautobot_config.py
@@ -67,51 +67,36 @@ DATABASES = {
     }
 }
 
-# Nautobot uses RQ for task scheduling. These are the following defaults.
-# For detailed configuration see: https://github.com/rq/django-rq#installation
-RQ_QUEUES = {
+# Redis variables
+REDIS_HOST = os.getenv("REDIS_HOST", "localhost")
+REDIS_PORT = os.getenv("REDIS_PORT", 6379)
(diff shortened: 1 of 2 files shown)


development/nautobot_config.py
```diff
@@ -86,6 +86,22 @@ def is_truthy(arg):
...
+    },
+    "custom_fields": {
+        "HOST": os.environ["REDIS_HOST"],
+        "PORT": int(os.environ.get("REDIS_PORT", 6379)),
+        "DB": 0,
+        "PASSWORD": os.environ["REDIS_PASSWORD"],
+        "SSL": is_truthy(os.environ.get("REDIS_SSL", False)),
+        "DEFAULT_TIMEOUT": 300,
+    },
 }
```
glennmatthews, line 105: Take a look at the updated `CACHES` settings that were made in https://github.com/nautobot/nautobot/pull/319 - you should be able to prune out a lot of this repeated boilerplate.

glennmatthews (commented):
Thanks! Do you want to update `pyproject.toml` and/or `__init__.py` with a minimum supported Nautobot version as well, or is this plugin still supporting all the way back to alpha?

development/nautobot_config.py
jvanderaa: Good call. Updates made, now just to get the tests to pass.

glennmatthews (approved):

PR #44, opened 2021-05-12
Update pyjwt version

matt852:
This updates the version of pyjwt from 1.7.1 to 2.1.0, along with any required dependencies.

I manually added in "cryptography" to pyproject.toml, as it's required for this and a few other packages.

Lastly, minor version bump from 1.2.0 to 1.3.0 as the backend dependencies changed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-chatops"
-version = "1.2.0"
+version = "1.3.0"
 description = "A plugin providing chatbot capabilities for Nautobot"
 authors = ["Network to Code, LLC <opensource@networktocode.com>"]
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


matt852:
I manually tested ChatOps functionality within MS Teams and Slack after the update, and found no issues.

glennmatthews (approved):
Looks reasonable to me - if it all works, :shipit:

jvanderaa (approved):

pyproject.toml
```diff
@@ -21,9 +21,10 @@ python = "^3.6"
 nautobot = "*"
 nautobot-capacity-metrics = "*"
 texttable = "^1.6.2"
-pyjwt = {extras = ["crypto"], version = "^1.7.1"}
+PyJWT = "^2.1.0"
 webexteamssdk = "^1.3"
 slack-sdk = "^3.4.2"
+cryptography = "^3.4.7"
```
dgarros, line 27: this library is used by a lot of other project, might be better to not pin its version to specific version but rather keep it flexible (if possible)

dgarros: I should have pushed harder on this one, I'm blocked because of this issue right now :(

PR #46, opened 2021-05-14
Update README.md to highlight demos and screen shots

jedelman8:
Adds screen shots and mention of Slack so users can try out the chatbot.


jvanderaa (approved):
LG

glennmatthews:
Shouldn't the images be committed to the repository and linked from there?

PR #49, opened 2021-05-27
Add API for CommandTokens.

### LINKED ISSUE #48: Add API for Command Tokens
### Environment
* Nautobot version:  1.0.1
* nautobot-plugin-chatops version:  1.3.0

### Proposed Functionality

CommandTokens were added in to support Multiple commands in Mattermost. At the time this was implemented, an API was not added. 

### Use Case

With the extensibility of Nautobot-plugin-chatops, a user may need add multiple commands for Mattermost. This allows for easy management of the tokens required for the Mattermost integration.

smk4664:
Fixes #48

I had to update to Nautobot 1.0.1 to get the API test cases to work.


pyproject.toml
```diff
@@ -27,7 +27,7 @@ slack-sdk = "^3.4.2"
 cryptography = "^3.4.7"
 
 [tool.poetry.dev-dependencies]
-black = "*"
+black = "^20.8b1"
```
jvanderaa, line 30:
```suggestion
black = "*"
```

Leaving at * as the poetry lock file handles the versioning.

jvanderaa (changes requested):

jvanderaa:
Does that mean we need to put a requirement of Nautobot 1.0.1 for this? I don't recall seeing the minimum version change.

pyproject.toml
smk4664: That's the problem I had... running Poetry lock did not handle the "*".

smk4664:
> Does that mean we need to put a requirement of Nautobot 1.0.1 for this? I don't recall seeing the minimum version change.

That's why I called this out.  I wasn't sure if I needed to do it here, or when a release was created.

pyproject.toml
smk4664: nautobot-plugin-chatops % poetry lock
The lock file might not be compatible with the current version of Poetry.
Upgrade Poetry to ensure the lock file is read properly or, alternatively, regenerate the lock file with the `poetry lock` command.
Updating dependencies
Resolving dependencies... (0.8s)
[SolverProblemError]
Because nautobot-chatops depends on black (*) which doesn't match any versions, version solving failed.

Same error when running `Poetry update`.

smk4664:
I will say, the endpoint will work on older versions of Nautobot, just not the tests. The tests require 1.0.1.

pyproject.toml
glennmatthews: Can you check what version of Poetry you have?

glennmatthews (commented):
I'll suggest again that plugins shouldn't actually declare (in pyproject.toml/poetry.lock) a dependency on Nautobot itself, as it can result in the situation you encountered where installing the plugin inadvertently overwrites an existing Nautobot installation with a different version than what you were expecting.

- For Docker dev environments, use the correct `networktocode/nautobot-dev` image as a base image so that Nautobot comes pre-installed (saving you build time as a bonus)
- For local/virtualenv dev environments, document that the developer is responsible for doing a `pip install nautobot==$NAUTOBOT_VER` with whatever version is desired.

It's a bit inelegant but I think it makes for less pain in the long term.

pyproject.toml
smk4664: I thought I had an updated version, but turns out I don't. I will try this again.

smk4664:
Thanks for the feedback @glennmatthews. That makes sense to me. I will remove the changes to the lock file and update the documentation.

nautobot_chatops/tests/test_api.py
```diff
@@ -0,0 +1,37 @@
...
+        {"comment": "Test 6", "platform": "mattermost", "token": "token6"},
+    ]
+    bulk_update_data = {"comment": "Testing"}
+
+    @classmethod
+    def setUpTestData(cls):
+        """Generate test data for the CommandToken Endpoint."""
+        CommandToken.objects.create(comment="Test 1", platform="mattermost", token="token1")
+        CommandToken.objects.create(comment="Test 2", platform="mattermost", token="token2")
+        CommandToken.objects.create(comment="Test 3", platform="mattermost", token="token3")
```
jvanderaa, line 37: Are there any tests that need to be added for this?

nautobot_chatops/__init__.py
```diff
@@ -40,7 +40,7 @@ class NautobotChatOpsConfig(PluginConfig):
         "mattermost_url": None,
     }
     # TODO dgarros : Define min and max version when things stabilize
```
glennmatthews, line 42: Probably can remove the TODO and set a `max_version = "1.999"?

glennmatthews (approved):

nautobot_chatops/__init__.py
```diff
@@ -39,7 +39,8 @@ class NautobotChatOpsConfig(PluginConfig):
         "mattermost_api_token": None,
         "mattermost_url": None,
     }
-    # TODO dgarros : Define min and max version when things stabilize
+
+    max_version = "1.999"
     min_version = "1.0.0"
     # max_version = ""
```
glennmatthews, line 45:
```suggestion
```

pyproject.toml
```diff
@@ -90,70 +89,3 @@ max-line-length=120
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
-
-[tool.pylint.master]
```
glennmatthews: Intentional to remove all of the below?

tasks.py
```diff
@@ -376,8 +376,6 @@ def tests(context, failfast=False):
     pydocstyle(context)
     print("Running pylint...")
     pylint(context)
-    print("Running yamllint...")
-    yamllint(context)
```
glennmatthews: Intentional removal?

smk4664: No... got lost... One minute.

pyproject.toml
smk4664: There were duplicate keys.  If you recheck the file, it should all be in there.  That is why a previous commit or two failed.  Git really let me down today.

glennmatthews (approved):

jvanderaa (approved):

PR #51, opened 2021-05-30
pylint cleanup

tanvimoharir:
- [x] Removing the global suppressions from the .pylintrc and fixing them in the respective modules or disabling them in the respective modules.
- [x] Moving configurations to pyproject.toml


development/nautobot_config.py
```diff
@@ -7,6 +7,8 @@
 from django.core.exceptions import ImproperlyConfigured
 from nautobot.core.settings import *  # noqa: F401,F403 pylint: disable=wildcard-import,unused-wildcard-import
 
+# pylint: disable=line-too-long
```
glennmatthews, line 10: Is this still needed given that you're setting `max-line-length=120` in the config?

nautobot_chatops/api/views/mattermost.py
```diff
@@ -36,7 +37,7 @@ def verify_signature(request):
         # in these cases, we have to load the request.body
         try:
             data = json.loads(request.body)
-        except ValueError as e:
+        except ValueError as e:  # pylint:disable=invalid-name
```
glennmatthews, line 40: It'd be better to just change `e` to something like `exc` throughout, as that would stop pylint from complaining.

.pylintrc
```diff
@@ -14,38 +14,7 @@ notes=FIXME,
 # Black handles line-too-long for us
```
glennmatthews, line 14: If things are added correctly to the `pyproject.toml`, it should be possible to remove the `.pylintrc` file entirely. You can refer to https://github.com/nautobot/nautobot-plugin-netbox-importer/blob/develop/pyproject.toml#L59 as an example.

nautobot_chatops/dispatchers/base.py
```diff
@@ -70,6 +70,7 @@ def platform_lookup(cls, item_type, item_name):
         """
         raise NotImplementedError
 
+    # pylint: disable=no-else-return
```
glennmatthews, line 73: This one could probably be fixed by simply removing the `else`, i.e.:

```python
    def static_url(self, path):
        """Construct an absolute URL for the given static file path, such as "nautobot/NautobotLogoSquare.png"."""
        static_path = str(static(path))
        if static_path.startswith("http"):
            return static_path
        return f"{self.context['request_scheme']}://{self.context['request_host']}{static_path}"
```

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -574,12 +577,14 @@ def bold(self, text):
         """Mark text as bold."""
         return f"**{text}**"
 
+    # pylint: disable=arguments-differ
```
glennmatthews, line 580: This one worries me - what arguments are differing?

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -601,6 +606,7 @@ def markdown_element(self, text):
         """Construct a basic Markdown-formatted text element."""
         return text
 
+    # pylint: disable=arguments-differ
```
glennmatthews, line 609: This one as well - can it be fixed instead of disabled?

nautobot_chatops/tables.py
```diff
@@ -34,6 +35,7 @@ class CommandLogTable(BaseTable):
         template_code='<span class="label label-{{record.status_label_class}}">{{ record.status }}</span>'
     )
 
+    # pylint: disable=too-few-public-methods
```
glennmatthews, line 38: I'd be fine with disabling `too-few-public-methods` globally in `pyproject.toml` instead of inline here.

glennmatthews (commented):
Thank you for taking this on!

development/nautobot_config.py
tanvimoharir: This is for lines which are greater than 120.

nautobot_chatops/api/views/mattermost.py
tanvimoharir: Sure,will do.

development/nautobot_config.py
glennmatthews: Thanks. How many such lines are there? Can they be easily reformatted to fit the 120-character length instead?

nautobot_chatops/tables.py
tanvimoharir: Done

nautobot_chatops/dispatchers/mattermost.py
tanvimoharir: The message was something like this:
W0221: Parameters differ from overridden 'send_blocks' method (arguments-differ)

nautobot_chatops/dispatchers/base.py
tanvimoharir: done

.pylintrc
tanvimoharir: Thanks for this.I have removed the .pylintrc file now.

development/nautobot_config.py
tanvimoharir: There were few,tried to resolve most of them there was one case wherein when I tried to break the line black tried to change it back.

nautobot_chatops/dispatchers/slack.py
```diff
@@ -297,7 +297,8 @@ def get_prompt_from_menu_choices(self, choices, offset=0):
             try:
                 # Since we are showing "Next..." at the end, this isn't required to show to users anymore
                 self.send_warning(
-                    f"More than {self.slack_menu_limit} options are available. Slack limits us to only displaying {self.slack_menu_limit} options at a time."
+                    f"More than {self.slack_menu_limit} options are available. \
+                     Slack limits us to only displaying {self.slack_menu_limit} options at a time."
```
glennmatthews, line 301: I think the below would work as well and would avoid using backslash-newline?

```suggestion
                    f"More than {self.slack_menu_limit} options are available. "
                    f"Slack limits us to only displaying {self.slack_menu_limit} options at a time."
```

nautobot_chatops/workers/nautobot.py
```diff
@@ -540,7 +542,8 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
             "nautobot",
             "get-interface-connections",
             [("Filter type", filter_type), ("Filter value 1", filter_value_1)],
-            "Interface Connection List, Note that a `(*)` indicates the device isn’t in the grouping requested, but is connected to a device that is.",
+            "Interface Connection List, Note that a `(*)` indicates the device isn’t in \
+             the grouping requested, but is connected to a device that is.",
```
glennmatthews, line 546: Similarly, perhaps:

```suggestion
            "Interface Connection List, Note that a `(*)` indicates the device isn’t in "
            "the grouping requested, but is connected to a device that is.",
```

nautobot_chatops/workers/nautobot.py
```diff
@@ -630,7 +633,8 @@ def change_device_status(dispatcher, device_name, status):
                 nautobot_logo(dispatcher),
             ),
             dispatcher.markdown_block(
-                f"Nautobot status for {dispatcher.bold(device_name)} successfully changed to {dispatcher.monospace(status)}."
+                f"Nautobot status for {dispatcher.bold(device_name)} \
+                successfully changed to {dispatcher.monospace(status)}."
```
glennmatthews, line 637:
```suggestion
                f"Nautobot status for {dispatcher.bold(device_name)} "
                f"successfully changed to {dispatcher.monospace(status)}."
```

tasks.py
```diff
@@ -75,7 +74,8 @@ def generate_packages(context, nautobot_ver=NAUTOBOT_VER, python_ver=PYTHON_VER)
         pty=True,
     )
     context.run(
-        f"docker-compose  -f {COMPOSE_FILE} -p {BUILD_NAME} run --name {container_name} -w /source nautobot poetry build",
+        f"docker-compose  -f {COMPOSE_FILE} -p {BUILD_NAME} run --name {container_name} \
+        -w /source nautobot poetry build",
```
glennmatthews, line 78:
```suggestion
        f"docker-compose  -f {COMPOSE_FILE} -p {BUILD_NAME} run --name {container_name} "
        "-w /source nautobot poetry build",
```

tasks.py
```diff
@@ -222,7 +222,8 @@ def create_user(context, user="admin", nautobot_ver=NAUTOBOT_VER, python_ver=PYT
     DEFAULT_ENV["PYTHON_VER"] = python_ver
 
     context.run(
-        f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot nautobot-server createsuperuser --username {user}",
+        f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot nautobot-server createsuperuser \
+         --username {user}",
```
glennmatthews, line 226:
```suggestion
        f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot "
        f"nautobot-server createsuperuser --username {user}",
```

Though in the long term we should refactor the `tasks.py` to match the newer pattern we have in other plugins (e.g. https://github.com/nautobot/nautobot-plugin-netbox-importer/blob/develop/tasks.py, https://github.com/nautobot/nautobot-plugin-circuit-maintenance/blob/develop/tasks.py) where we use some helper functions to reduce the need to state `docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot` all over the place.

tasks.py
```diff
@@ -248,7 +249,8 @@ def makemigrations(context, name="", nautobot_ver=NAUTOBOT_VER, python_ver=PYTHO
 
     if name:
         context.run(
-            f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot nautobot-server makemigrations --name {name}",
+            f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot nautobot-server makemigrations \
+            --name {name}",
```
glennmatthews, line 253:
```suggestion
            f"docker-compose -f {COMPOSE_FILE} -p {BUILD_NAME} run nautobot "
            f"nautobot-server makemigrations --name {name}",
```

glennmatthews (approved):
Nice work!

nautobot_chatops/dispatchers/slack.py
tanvimoharir: Makes sense, I will incorporate these as well.

tanvimoharir: This is done

nautobot_chatops/workers/nautobot.py
tanvimoharir: This is done

tasks.py
tanvimoharir: This is done.

tasks.py
tanvimoharir: This is done

Issue #45, opened 2021-05-12
Unpin Cryptography


jvanderaa:
### Environment
* nautobot-plugin-chatops version: 1.3.0

### Proposed Functionality

Unpin specific version, go to a less specific version requirement for `cryptography`, to version 3.x.x

### Use Case

Allow for the updating of cryptography

PR #54, opened 2021-06-10
Remove cryptography as a top level dependency

dgarros:
Fixes : #45 

This PR removes `cryptography` from the list of top level dependency,
As far as I can tell we are not using cryptography directly in the chatops so it probably shouldn't be a direct dependency.


dgarros:
I also noticed that the `[build-system]` of the pyproject.toml was not up to date so I've updated it

matt852:
LGTM.  I recommend @jvanderaa review this as we both worked on the original PR that introduced it.

jvanderaa (approved):

Issue #24, opened 2021-03-17
Rename Webex Teams -> Webex


jvanderaa:
### Environment
* Python version:  N/A
* Nautobot version:  N/A
* nautobot-plugin-chatops version: 1.0.0

### Observed Behavior

Webex Teams has been renamed to just **Webex**. This may help the messaging to rename webex_teams work to just webex where applicable.

matt852:
Please assign this to me.  I submitted PR #55 to address this issue.

jvanderaa:
Closed by #55

PR #55, opened 2021-06-11
#24 - Rename Webex Teams to Webex

matt852:
Addresses issue #24 by renaming all instances of "WebEx Teams" to just "WebEx".


matt852:
This change will require the environment variables for Webex to be updated, as well as any existing webhooks to be updated with the new URL.  All of these changes are updated in the docs.

nautobot_chatops/api/urls.py
```diff
@@ -23,11 +23,11 @@
         path("ms_teams/messages/", MSTeamsMessagesView.as_view(), name="ms_teams_messages"),
     ]
 
-if settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_webex_teams"):
-    from nautobot_chatops.api.views.webex_teams import WebExTeamsView
+if settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_webex"):
```
glennmatthews, line 26: Can we make this backwards-compatible for a release or two? Something like:

```suggestion
if (
    settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_webex") or 
    settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_webex_teams")
):
    if (
        "enable_webex_teams" in settings.PLUGINS_CONFIG["nautobot_chatops"] and
        "enable_webex" not in settings.PLUGINS_CONFIG["nautobot_chatops"]
    ):
        logger.warning("The 'enable_webex_teams' setting is deprecated. Please use 'enable_webex' instead.")
```

(You'll of course need to add `import logging` and `logger = logging.getLogger(__name__)` at the head of this file.)

nautobot_chatops/api/urls.py
```diff
@@ -23,11 +23,11 @@
...
+    from nautobot_chatops.api.views.webex import WebExView
 
     urlpatterns += [
-        path("webex_teams/", WebExTeamsView.as_view(), name="webex_teams"),
+        path("webex/", WebExView.as_view(), name="webex"),
```
glennmatthews, line 30: Similarly, for backwards compatibility, perhaps support both temporarily?

```suggestion
        path("webex/", WebExView.as_view(), name="webex"),
        path("webex_teams/", WebExView.as_view(), name="webex_teams"),
```

nautobot_chatops/api/views/webex.py
```diff
@@ -16,20 +16,20 @@
...
 
 from nautobot_chatops.workers import get_commands_registry, commands_help, parse_command_string
-from nautobot_chatops.dispatchers.webex_teams import WebExTeamsDispatcher
+from nautobot_chatops.dispatchers.webex import WebExDispatcher
 from nautobot_chatops.utils import check_and_enqueue_command
 
 logger = logging.getLogger(__name__)
 
-TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
+TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
```
glennmatthews, line 24:
```suggestion
if "webex_token" in settings.PLUGINS_CONFIG["nautobot_chatops"]:
    TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
else:
    TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
    logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead")
```

nautobot_chatops/api/views/webex.py
```diff
@@ -39,12 +39,12 @@
 
 def generate_signature(request):
     """Calculate the expected signature of a given request."""
-    signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams_signing_secret").encode("utf-8")
+    signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_signing_secret").encode("utf-8")
```
glennmatthews, line 42:
```suggestion
    if ["webex_signing_secret"] in settings.PLUGINS_CONFIG["nautobot_chatops"]:
        signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_signing_secret").encode("utf-8")
    else:
        signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams_signing_secret").encode("utf-8")
        logger.warning("The 'webex_teams_signing_secret' setting is deprecated, please use 'webex_signing_secret' instead")
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -44,8 +44,8 @@ def subclasses(cls):
         if settings.PLUGINS_CONFIG["nautobot_chatops"].get("ms_teams"):
             from .ms_teams import MSTeamsDispatcher
 
-        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams"):
-            from .webex_teams import WebExTeamsDispatcher
+        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex"):
+            from .webex import WebExDispatcher
```
glennmatthews, line 48: How has this ever worked?? :-)

Need to change these to `"enable_ms_teams"` and `"enable_webex"` respectively.

nautobot_chatops/dispatchers/webex.py
```diff
@@ -13,26 +13,26 @@
...
+    platform_slug = "webex"
 
     platform_color = "6EBE4A"
 
     def __init__(self, *args, **kwargs):
-        """Init a WebExTeamsDispatcher."""
+        """Init a WebExDispatcher."""
         super().__init__(*args, **kwargs)
-        self.client = WebexTeamsAPI(access_token=settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"])
+        self.client = WebexTeamsAPI(access_token=settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"])
```
glennmatthews, line 35:
```suggestion
        if "webex_token" in settings.PLUGINS_CONFIG["nautobot_chatops"]:
            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
        else:
            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
            logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead.")
        self.client = WebexTeamsAPI(access_token=access_token)
```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-chatops"
-version = "1.3.1"
+version = "1.4.0"
```
glennmatthews, line 3: As an alternative to backwards-compatibility as I'm suggesting in my other changes, if you don't want to maintain backwards compatibility, this would need to become `2.0.0` per semantic versioning.

glennmatthews (commented):
I don't know how many WebEx users we have for this plugin, but I don't think we should break backwards-compatibility lightly.

jvanderaa:
I'd like to have some of the backwards compatibility methodology if modifying the code. The first part is the front side of things and how it is publicly displayed versus fixing the back end components. Then start that migration over time to the new name.

pyproject.toml
matt852: I like your thoughts on backwards compatibility, so I'll leave the version as is.  I think we can make 2.0.0, whenever that is, the change that breaks backwards compatibility for Webex.

matt852:
The points on leaving backwards compatibility are strong, so I'll go back and add it in.  Great feedback @glennmatthews @jvanderaa

nautobot_chatops/api/urls.py
matt852: Good idea.  Updated.

nautobot_chatops/api/urls.py
matt852: Updated

nautobot_chatops/api/views/webex.py
matt852: Updated

nautobot_chatops/api/views/webex.py
matt852: Updated

nautobot_chatops/dispatchers/webex.py
matt852: Updated

nautobot_chatops/dispatchers/base.py
matt852: This entire `subclasses` function doesn't appear to be called at all during normal ChatOps interactions.  I would suggest holding off on changing these for now, and rewriting them in a separate PR (or remove them completely if found to not be needed).
@glennmatthews What do you think?

glennmatthews: Looks like it's mainly used by the `AccessLookupView` API view class, which is used in in the AccessGrant create/edit views to facilitate looking up the user/channel ID based on the user/channel name. They're clearly wrong as implemented so I'd prefer to fix these as part of this PR as long as you're touching this code anyways.

docs/chat_setup/webex_setup.md
```diff
@@ -21,13 +21,22 @@
...
-4. Configure the `webex_teams_signing_secret` in your `.creds.env` to match the secret string that you selected in
+4. Configure the `webex_signing_secret` in your `.creds.env` to match the secret string that you selected in
    step 3 above.
 5. Proceed to the [Nautobot Server Preparation and Configuration](./chat_setup.md#Nautobot Server Preparation and Configuration) section.
 
+## Deprecation Warning
+
+As of Nautobot ChatOps Plugin v1.4.0, the PLUGIN_CONFIG settings for Webex has changed to align with the official renaming of `Webex Teams` to `Webex`:
+
+- `webex_teams_token` is deprecated. Use `webex_token` instead.
```
glennmatthews, line 37: What about `enable_webex_teams`/`enable_webex`?

nautobot_chatops/dispatchers/webex.py
```diff
@@ -13,26 +13,32 @@
...
-        """Init a WebExTeamsDispatcher."""
+        """Init a WebExDispatcher."""
         super().__init__(*args, **kwargs)
-        self.client = WebexTeamsAPI(access_token=settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"])
+        # v1.4.0 Deprecation warning
+        if "webex_token" in settings.PLUGINS_CONFIG["nautobot_chatops"]:
+            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
+        else:
+            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
+            logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead.")
```
glennmatthews, line 40:
```suggestion
        if (
            "webex_teams_token" in settings.PLUGINS_CONFIG["nautobot_chatops"]
            and "webex_token" not in settings.PLUGINS_CONFIG["nautobot_chatops"]
        ):
            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
            logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead.")
        else:
            access_token = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
```

docs/chat_setup/webex_setup.md
matt852: My mistake.  Updated.

nautobot_chatops/api/views/webex.py
```diff
@@ -16,20 +16,25 @@
...
 
 logger = logging.getLogger(__name__)
 
-TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
+# v1.4.0 Deprecation warning
+if "webex_token" in settings.PLUGINS_CONFIG["nautobot_chatops"]:
+    TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
+else:
+    TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
+    logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead")
```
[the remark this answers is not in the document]
matt852: I think this is a good suggestion. I've added it and updated it slightly to explicitly handle the exception and log it accordingly.

nautobot_chatops/api/views/webex.py
```diff
@@ -39,12 +44,19 @@
...
     """Calculate the expected signature of a given request."""
-    signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams_signing_secret").encode("utf-8")
+    # v1.4.0 Deprecation warning
+    if "webex_signing_secret" in settings.PLUGINS_CONFIG["nautobot_chatops"]:
+        signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_signing_secret").encode("utf-8")
+    else:
+        signing_secret = settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams_signing_secret").encode("utf-8")
+        logger.warning(
+            "The 'webex_teams_signing_secret' setting is deprecated, please use 'webex_signing_secret' instead"
+        )
```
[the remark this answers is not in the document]
matt852: Added this in with similar exception handling added like in the previous suggestion.

nautobot_chatops/dispatchers/webex.py
matt852: Added this in with similar exception handling added like in the previous suggestions.

nautobot_chatops/dispatchers/base.py
matt852: Updated/fixed

glennmatthews (approved):

matt852:
@jvanderaa All tests are passing locally.  TravisCI was throwing this error: ```toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit```

nautobot_chatops/dispatchers/base.py
```diff
@@ -41,11 +41,14 @@ def subclasses(cls):
...
             from .slack import SlackDispatcher
 
-        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("ms_teams"):
+        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_ms_teams"):
             from .ms_teams import MSTeamsDispatcher
 
-        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("webex_teams"):
-            from .webex_teams import WebExTeamsDispatcher
+        if settings.PLUGINS_CONFIG["nautobot_chatops"].get("enable_webex"):
+            from .webex import WebExDispatcher
```
jvanderaa, line 48: Do we also need a `enable_webex_teams` section here?

nautobot_chatops/choices.py
```diff
@@ -41,6 +41,6 @@ class CommandTokenPlatformChoices(ChoiceSet):
     MATTERMOST = "mattermost"
     SLACK = "slack"
     MS_TEAMS = "microsoft_teams"
-    WEBEX_TEAMS = "webex_teams"
+    WEBEX = "webex"
```
jvanderaa, line 44: Should we keep WEBEX_TEAMS for this particular setup?

jvanderaa (commented):
A couple of questions within the backwards compatibility section. Let me know your thoughts, and may want @glennmatthews thoughts there too on the two comments.

nautobot_chatops/dispatchers/base.py
matt852: That's a good point I didn't catch.  Updated

nautobot_chatops/choices.py
matt852: Currently only Mattermost is being used, so this is just forward looking.
As for keeping it, I'm not sure why we'd want to when everything else is changing to just WEBEX.  Is there any specific scenario you had in mind for keeping it as WEBEX_TEAMS?

nautobot_chatops/api/views/webex.py
```diff
@@ -16,20 +16,34 @@
...
+):
+    TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"]
+    logger.warning("The 'webex_teams_token' setting is deprecated, please use 'webex_token' instead")
+else:
+    try:
+        TOKEN = settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_token"]
+    except KeyError as err:
+        ERROR_MSG = "The 'webex_token' setting must be configured"
+        logger.error(ERROR_MSG)
+        raise KeyError(ERROR_MSG) from err
```
smk4664, line 37: I am concerned about this, as it seems like this would crash the chatops plugin. This is different than the previous that would only warn if the token was missing. This wouldn't matter if webex was disabled as this file is still loaded.

nautobot_chatops/dispatchers/webex.py
```diff
@@ -13,26 +13,42 @@
...
+    platform_slug = "webex"
 
     platform_color = "6EBE4A"
 
     def __init__(self, *args, **kwargs):
-        """Init a WebExTeamsDispatcher."""
+        """Init a WebExDispatcher."""
         super().__init__(*args, **kwargs)
-        self.client = WebexTeamsAPI(access_token=settings.PLUGINS_CONFIG["nautobot_chatops"]["webex_teams_token"])
+        # v1.4.0 Deprecation warning
```
smk4664, line 35: This is duplicated code, you might want to make a function for it.

nautobot_chatops/api/views/webex.py
smk4664: In testing, this didn't seem to cause the error like I thought, but it is different than the previous code that just emitted a warning.  I think this is due to the default_setting being `None`.  Which makes me wonder if this code is even going to be executed.

matt852: I believe even if this threw an error, we would want it to crash.  The Try/Except statement is to log it properly before crashing.
It's saying if the Webex client is being loaded, but no token is set for either the old name or new name, log the error and fail out.

nautobot_chatops/dispatchers/webex.py
matt852: This is in a couple of places across the webex files when initially loading the token when initializing the class, but I think since this is just for a deprecation warning where it will be removed at a later date, it should be ok to leave as is.

nautobot_chatops/api/views/webex.py
smk4664: I am not a huge fan of letting an app crash as far as user experience goes.  But, this code will never be executed due to the plugin defaults.

nautobot_chatops/choices.py
smk4664: Webex is good hear, that is correct it is not used in code yet, but I have plans for it.

smk4664 (approved):
Thanks for testing and making these changes!

matt852:
@jvanderaa @glennmatthews I believe this is ready to merge in.

jvanderaa (approved):

jvanderaa:
Thank you @matt852 & @smk4664 !!

PR #57, opened 2021-06-16
a few documentation cleanups

tim-fiola:
Added a line to do a nautobot server migration after editing the nautobot_config.py file as I've had to do that each time in order to get it to work

### DIFF
--- a/docs/chat_setup/chat_setup.md
+++ b/docs/chat_setup/chat_setup.md
@@ -88,6 +88,7 @@ PLUGINS_CONFIG = {
     }
 }
 ```
+After editing the `nautobot_config.py` file, as the `nautobot` user, run the `nautobot-server migrate` command.
 
 As a sudo-enabled user, restart the `nautobot` and `nautobot-worker` process after updating `nautobot_config.py`.
 


smk4664:
Editing the Nautobot_config.py should not require running the migrate command.  I am interested if something else is happening when you run the migrate commad?  It should just take a restart of the Worker and the Server.

docs/chat_setup/chat_setup.md
```diff
@@ -88,6 +88,7 @@ PLUGINS_CONFIG = {
     }
 }
 ```
+After editing the `nautobot_config.py` file, as the `nautobot` user, run the `nautobot-server migrate` command.
 
 As a sudo-enabled user, restart the `nautobot` and `nautobot-worker` process after updating `nautobot_config.py`.
```
jvanderaa, line 93: Combine into one if we believe the update is needed. I see this as more explicit than previously which is good, but we ask for the restart process twice.

jvanderaa (changes requested):
Thank you for the update!

tim-fiola:
Sorry folks. I should have opened this as a bug instead of a doc change.  Let me close this out and document what I see when I don't do the migration.

PR #61, opened 2021-07-01
Correct bug in Mattermost not sending interactive dialog.

### LINKED ISSUE #5: Mattermost dialog box doesn't auto-appear on interaction calls
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

In Mattermost, when I run the command `/nautobot get-devices` command, and select "Name" from the Select a Device Filter selection menu, no dialog box appears.

### Steps to Reproduce
1.  In Mattermost, run the command `/nautobot get-devices`
2. In the Select a Device Filter menu, select "Name"

### Expected Behavior

A dialog box to appear, allowing me to type in the device name.

### Observed Behavior

I got a private message from the bot saying "If the dialog does not appear, copy this command and resend: /nautobot get-devices name"

Nothing else appeared after that.  However if I run that exact command listed, `/nautobot get-devices name`, a dialog box properly appears and I'm able to pull up a device.

smk4664:
Fixes #5 

Mattermost expects to receive a json response to the  Interactive message.  Docs from mattermost show that you *should* respond with a message but in testing, the empty json gets accepted.   Responding with a message would impact every command or interaction with mattermost, which would be very annoying for the user.

### DIFF
--- a/nautobot_chatops/dispatchers/mattermost.py
+++ b/nautobot_chatops/dispatchers/mattermost.py
@@ -406,13 +406,6 @@ class MattermostDispatcher(Dispatcher):  # pylint: disable=too-many-public-metho
                     },
                     dialog_url=self.context.get("integration_url"),
                 )
-                # This is added as select menus are not issueing a valid trigger_id or the trigger_id is expired (default 3 second.)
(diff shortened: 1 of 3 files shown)


glennmatthews (approved):
Looks reasonable to me!

jvanderaa (approved):

PR #71, opened 2021-08-05
Fix version

jvanderaa:
Using the pyproject.toml to update the plugin version. Without this a manual step to update the version number is required.

### DIFF
--- a/nautobot_chatops/__init__.py
+++ b/nautobot_chatops/__init__.py
@@ -1,6 +1,7 @@
 """Nautobot plugin implementing a chatbot."""
-__version__ = "1.1.0"
+import toml
 
+__version__ = toml.load("pyproject.toml")["tool"]["poetry"]["version"]
 
 from nautobot.extras.plugins import PluginConfig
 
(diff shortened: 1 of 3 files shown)


nautobot_chatops/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Nautobot plugin implementing a chatbot."""
-__version__ = "1.1.0"
+import toml
 
+__version__ = toml.load("pyproject.toml")["tool"]["poetry"]["version"]
```
glennmatthews, line 4: I don't recall if this formulation works in a built wheel - I'd recommend instead the `importlib` approach we're using in nautobot itself (https://github.com/nautobot/nautobot/blob/develop/nautobot/__init__.py#L3) as well as other plugins (https://github.com/nautobot/nautobot-plugin-netbox-importer/blob/develop/nautobot_netbox_importer/__init__.py#L3, https://github.com/nautobot/nautobot-plugin-ssot/blob/develop/nautobot_ssot/__init__.py#L3)

jvanderaa: Good call. I'll make that update. I'll also update cookiecutter as well

glennmatthews (approved):

Issue #72, opened 2021-08-09
Allow for optional dialog box inputs


matt852:
### Environment
* Nautobot version:   1.0.2
* nautobot-plugin-chatops version:   1.4.0

### Proposed Functionality

When creating new chatops plugins or commands, allow an input from a dialog box to be optional.

### Use Case

I want to create a ChatOps command to trace a path between two IP addresses, where the Ports and Protocols are optional.  If provided, additional firewall checks are ran.

matt852:
Please assign this to me as I'm working on a PR for this.

PR #73, opened 2021-08-09
Optional input dialog

matt852:
Fixes issue #72 

Slack and Mattermost both use the `multi_input_dialog` function to leverage dialog boxes (modals) with multiple inputs.
This PR adds support for both clients to make any of the supported inputs (currently free field text and dropdown select boxes) optional when creating custom commands/plugins.


glennmatthews (commented):
Code changes look good - it would be good to see some test case coverage for this functionality though.

glennmatthews (approved):

jvanderaa (approved):
LG

Issue #75, opened 2021-08-12
Support MySQL


smk4664:
### Proposed Functionality

Currently, Nautobot-plugin-chatops uses django.contrib.postgres.fields.ArrayField which is not compatible with MySQL Backed Database Engines.  I propose we convert this field to nautobot.utilities.fields.JSONArrayField.

### Use Case

Now that Nautobot supports MySQL, we should support MySQL as well to maintain compatibility.

smk4664:
I will be taking up this effort.

jvanderaa:
Thanks @smk4664

jvanderaa:
Closed via #76

PR #76, opened 2021-08-14
Support mysql

smk4664:
FIXES: #75 

This PR introduces support for MySQL by converting the ArrayField to JSONField.  These changes require a migration to be run after upgrading. Reversing the migrations is supported, and testing was done on both postgres and mysql.  

This does not include updating the Travis stages to include mysql testing.


nautobot_chatops/models.py
```diff
@@ -28,9 +27,16 @@ class CommandLog(BaseModel):
...
+
+    @staticmethod
+    def params_default():
+        """Provide default for the params field.
+
+        Per Django docs, the default must be a value or a callable.
+        """
+        return []
+
+    params = models.JSONField(default=params_default, help_text="user_input_parameters")
```
glennmatthews, line 39: Does `default=list` work without needing a custom `params_default` method?

smk4664: Maybe, I could be misreading:

The default can’t be a mutable object (model instance, list, set, etc.), as a reference to the same instance of that object would be used as the default value in all new model instances. Instead, wrap the desired default in a callable. For example, if you want to specify a default dict for JSONField, use a function:

glennmatthews: Right - you wouldn't want `default=[]` because `[]` is an instance of a list and it would be shared across all records. `default=list` should be fine because `list()` is a callable that returns a new instance of an empty list each time it is called.

smk4664: Great news! I will make the changes!

glennmatthews (approved):
Looks good to me at a glance. Just to confirm - you've tested upgrading an existing 1.4.1 PostgreSQL installation with existing CommandLog records to this new version and verified that the migrated records appear correctly in the UI?

jvanderaa (approved):

Issue #64, opened 2021-07-14
Migrate to Celery Worker


nniehoff:
### Environment
* Nautobot version:  1.1.0
* nautobot-plugin-chatops version:  1.x

### Proposed Functionality

The Chatops plugin should use Celery workers in Nautobot 1.1

### Use Case

RQ is being deprecated

jvanderaa:
Any immediate thoughts on how to migrate gracefully?

glennmatthews:
See https://github.com/nautobot/nautobot-plugin-device-onboarding/pull/18 for one possible approach.

jvanderaa:
Are there any settings that can be imported to check on the current running version?

glennmatthews:
Sure, `settings.VERSION` should have the Nautobot version.

smk4664:
I will be taking on this issue today.

PR #78, opened 2021-08-17
Add celery worker

smk4664:
FIXES: #64 

This PR adds support for Celery Worker.  Verified that backwards compatibility was maintained.  Confirmed with nautobot-plugin-chatops-ansible to ensure that none of the chat command plugins would need to be updated.

I decided to keep the development environment setup the way it is, to ensure changes going forward are still done with backwards compatibility (at least until Nautobot 2.0).  I am providing a docker-compose file with celery so that local testing could move over to that if the developer wishes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ python:
 
 env:
   - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.2"
+  - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.1.2"
 
 services:
   - "docker"
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ python:
   - 3.8
 
 # Testing 1.0.3 as it is still pre-Celery, but has the choices test fix nautobot#457
+# 1.0.2 would also work, it would just skip the choices test.
 env:
   - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.3"
   - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.1.2"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,9 @@ python:
   - 3.7
   - 3.8
 
+# Testing 1.0.3 as it is still pre-Celery, but has the choices test fix nautobot#457
 env:
-  - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.2"
+  - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.3"
   - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.1.2"
 
 services:
--- a/development/docker-compose.celery.yml
+++ b/development/docker-compose.celery.yml
@@ -15,4 +15,3 @@ services:
     volumes:
       - "./nautobot_config.py:/opt/nautobot/nautobot_config.py"
       - "../:/source"
-
--- a/development/docker-compose.celery.yml
+++ b/development/docker-compose.celery.yml
@@ -7,7 +7,7 @@ services:
       - "dev.env"
       - "creds.env"
     tty: true
-    entrypoint: "nautobot-server celery worker -B -l INFO"
+    entrypoint: "nautobot-server celery worker -l INFO"
     depends_on:
       - "nautobot"
     healthcheck:
--- a/development/docker-compose.celery.yml
+++ b/development/docker-compose.celery.yml
@@ -1,0 +1,18 @@
+---
+version: "3.4"
+services:
+  celery:
+    image: "nautobot-chatops-plugin/nautobot:${NAUTOBOT_VER}-py${PYTHON_VER}"
+    env_file:
+      - "dev.env"
+      - "creds.env"
+    tty: true
+    entrypoint: "nautobot-server celery worker -B -l INFO"
+    depends_on:
+      - "nautobot"
+    healthcheck:
+      disable: true
+    volumes:
+      - "./nautobot_config.py:/opt/nautobot/nautobot_config.py"
+      - "../:/source"
+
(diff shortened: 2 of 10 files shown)


smk4664:
I apologize for failing checks, I will be working to fix these.

development/docker-compose.celery.yml
```diff
@@ -0,0 +1,17 @@
+---
+version: "3.4"
+services:
+  celery:
+    image: "nautobot-chatops-plugin/nautobot:${NAUTOBOT_VER}-py${PYTHON_VER}"
+    env_file:
+      - "dev.env"
+      - "creds.env"
+    tty: true
+    entrypoint: "nautobot-server celery worker -B -l INFO"
```
glennmatthews, line 10: Nitpick: `-B` isn't actually needed (it causes the worker to self-host Celery Beat, but we're not actually using that feature in 1.1, and it'll be different in 1.2, see https://github.com/nautobot/nautobot/pull/805/files#diff-338f8adcdd0eb410e4af13c5ab840ae9026cdecefae4a0a421579675749b9c2eR37). We should change this in Nautobot core (where you presumably copied this from) as well - I'll get an issue opened.

.travis.yml
```diff
@@ -13,8 +13,10 @@ python:
   - 3.7
   - 3.8
 
+# Testing 1.0.3 as it is still pre-Celery, but has the choices test fix nautobot#457
 env:
-  - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.2"
+  - "INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER=1.0.3"
```
glennmatthews, line 18: So is the issue here that the plugin works fine with 1.0.2 but tests will fail with that version? Might be good to state that in the README or similar then.

development/docker-compose.celery.yml
glennmatthews: https://github.com/nautobot/nautobot/issues/825 FYI.

nautobot_chatops/utils.py
```diff
@@ -14,6 +14,63 @@
...
+        # Get dispatcher class from module
+        dispatcher_class = getattr(sys.modules[dispatcher_module], dispatcher_name)
+
+        return function(subcommand, params=params, dispatcher_class=dispatcher_class, context=context)
+
+    def enqueue_task(command, subcommand, params, dispatcher_class, context, function=None):
+        """Enqueue task with Celery worker."""
+        # Since celery can't deserialize class, we will discard function and
+        # send the command instead.
+        _ = function
```
glennmatthews, line 43: I assume the assignment to `_` is just to avoid a `pylint:disable=unused-argument` comment? Personally I'd prefer the comment approach versus having actual do-nothing code.

nautobot_chatops/utils.py
```diff
@@ -155,11 +212,12 @@ def check_and_enqueue_command(
...
     # If we made it here, we're permitted. Enqueue it for the worker
-    get_queue("default").enqueue(
-        registry[command]["function"],
+    enqueue_task(
+        command,
         subcommand,
         params=params,
         dispatcher_class=dispatcher_class,
         context=context,
+        function=registry[command]["function"],
```
glennmatthews, line 221: Alternate idea - use keyword-only arguments plus `**kwargs` to handle the difference between required parameters, i.e.:

```python
    enqueue_task(
        function=registry[command]["function"],
        command=command,
        subcommand=subcommand,
        params=params,
        dispatcher_class=dispatcher_class,
        context=context,
    )
```

```python
# Celery
def enqueue_task(*, command, subcommand, params, dispatcher_class, context, **kwargs):
```

```python
# RQ
def enqueue_task(*, function, subcommand, params, dispatcher_class, context, **kwargs):
```

development/docker-compose.celery.yml
smk4664: Thanks! I actually copied it from the onboarding plugin

.travis.yml
smk4664: They don't actually fail in 1.0.2 either, it just doesn't do the choices tests.  The Choices test was added in 1.0.3 and failing when I ran 1.1.2 because I was missing `choices_fields`

nautobot_chatops/utils.py
smk4664: Yeah, it was trying to avoid that.  I will change.

nautobot_chatops/utils.py
smk4664: This is the one thing I am not good at in Python is kwargs... So, I tend to avoid them.

nautobot_chatops/utils.py
```diff
@@ -14,6 +14,61 @@
...
+        function, registry or dispatcher_class, we will have to look them up.
+        """
+        # import done here instead of up top to prevent circular imports.
+        # Disabling cyclic-import as this does not affect the worker.
+        # pylint: disable=import-outside-toplevel,cyclic-import
+        from nautobot_chatops.workers import get_commands_registry
+
+        registry = get_commands_registry()
+        function = registry[command]["function"]
+        # Get dispatcher class from module
```
jvanderaa, line 34: Any further details here? I think I have an idea of what it is doing, but this may be something worth having additional comments on.

jvanderaa (commented):
I think looks good. We'll get Glenn's 👁️  on it though

docs/contributing.md
```diff
@@ -40,6 +40,8 @@ provided `creds.env.example` file to `creds.env` and customizing the contents of
 
 ### Testing
 
+While the plugin itself supports any general release version of Nautobot, the testing requires at minimum 1.0.2.
```
glennmatthews, line 43: Just double-checking - is "1.0.2" correct here or did you mean "1.0.3"? If the former, what's different in 1.0.1 that it would not support the tests?

glennmatthews (approved):

docs/contributing.md
smk4664: Well... Tests will fail if you use 1.0.0 or 1.0.1.  They will pass with 1.0.2, but the choices test won't run until 1.0.3.  So should I change minimum to 1.0.3?

glennmatthews: 1.0.2 sounds fine/correct to me then. Thanks! :)

jvanderaa (approved):

Issue #17, opened 2021-03-12
Error filtering interface connections by region


matt852:
### Environment
* Python version:   3.8.5
* Nautobot version:   1.0.0b2
* nautobot-plugin-chatops version:   1.0.1

There is an error across all supported chatops platforms if you try to filter by region for the command `/nautobot get-interface-connections`, but no regions are configured in Nautobot.  This also occurs if regions are configured, but no sites are assigned a region.

NOTE: If a region is configured, and a site is assigned to a region, this bug does _not_ trigger even if there are no interfaces connections in that region.

Each chatops program gives different responses:

- MS Teams - untested
- Mattermost - Nothing happens
- Webex Teams - this message appears:

`If you're seeing this message, something went wrong`

- Slack - this message appears:

```
Sorry <user>, an error occurred :sob:
The request to the Slack API failed.
The server responded with: {'ok': False, 'error': 'internal_error'}
```

### Steps to Reproduce
1. Either way will trigger the bug here:
    1a. In Nautobot, verify no regions are configured
    1b. In Nautobot, verify no sites are assigned to any configured regions (regions can still be configured to trigger this bug)
2. In any chatops program, run command `/nautobot get-interface-connections`
3. In the initial filter box labeled "Select an interface connection filter", select "Region"

### Expected Behavior

A message or similar stating no regions are available to filter by
OR
Region to not even be a filterable option

### Observed Behavior

A workflow crash occurs, with some platforms responding with an error message, and others do not (see above for platform specific details)

jvanderaa:
@matt852 is this still an issue?

matt852:
@jvanderaa Yes.  I haven't had a chance to look into this yet.

smk4664:
I will try to submit a PR for this soon.  I have an idea of how to return a good warning to the user.

PR #81, opened 2021-08-18
Provides a response with empty choices

jvanderaa:
- Adds a check for choices, if it is an empty list where the error on #17 was occurring, then it sends a message back to the individual, returning False.

### DIFF
--- a/nautobot_chatops/workers/nautobot.py
+++ b/nautobot_chatops/workers/nautobot.py
@@ -388,6 +388,14 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
                 f'Unknown filter type "{filter_type}"',
             )  # command did not run to completion and therefore should not be logged
 
+        # Check on empty choice list, send an error back
+        if not choices:
+            dispatcher.send_markdown(
+                message=f"Unable to apply filter on '{filter_type}', please validate that data is present for a filter",
+                ephemeral=True
+            )
+            return False
+
         if filter_type != "device":
             dispatcher.prompt_from_menu(
                 f"nautobot get-interface-connections {filter_type}",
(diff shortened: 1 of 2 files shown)


nautobot_chatops/workers/nautobot.py
```diff
@@ -388,6 +388,14 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
                 f'Unknown filter type "{filter_type}"',
             )  # command did not run to completion and therefore should not be logged
 
+        # Check on empty choice list, send an error back
+        if not choices:
+            dispatcher.send_markdown(
+                message=f"Unable to apply filter on '{filter_type}', please validate that data is present for a filter",
```
glennmatthews, line 394: Perhaps?

```suggestion
                message=f"Unable to filter by '{filter_type}', as it appears there is no corresponding data available",
```

nautobot_chatops/workers/nautobot.py
```diff
@@ -388,6 +388,14 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
...
+                ephemeral=True
+            )
+            return False
```
glennmatthews, line 397: This might be worth recording as a STATUS_FAILED case, as it shouldn't be happening once you have a well-populated Nautobot instance?

jvanderaa: I was 50/50 on this one. I think that does make sense.

nautobot_chatops/workers/nautobot.py
```diff
@@ -388,6 +388,14 @@ def get_interface_connections(dispatcher, filter_type, filter_value_1, filter_va
                 f'Unknown filter type "{filter_type}"',
             )  # command did not run to completion and therefore should not be logged
 
+        # Check on empty choice list, send an error back
+        if not choices:
+            dispatcher.send_markdown(
+                message=f"Unable to filter by '{filter_type}', as it appears there is no corresponding data available",
+                ephemeral=True,
+            )
+            return CommandStatusChoices.STATUS_FAILED
```
glennmatthews, line 397:
```suggestion
            return (
                CommandStatusChoices.STATUS_FAILED,
                f'No choices found when filtering by "{filter_type}"',
            )
```

glennmatthews (approved):

PR #82, opened 2021-08-19
Removes Postgres and Redis port exposure

jvanderaa:
- Found that the dev environment had Redis and Postgres ports exposed in the Docker-Compose, this is not necessary. Removing for future dev.

### DIFF
--- a/development/docker-compose.requirements.yml
+++ b/development/docker-compose.requirements.yml
@@ -8,8 +8,6 @@ services:
       - "creds.env"
     volumes:
       - "postgres_data:/var/lib/postgresql/data"
-    ports:
-      - "5432:5432"
   redis:
     image: "redis:6-alpine"
     command:
@@ -19,7 +17,5 @@ services:
     env_file:
       - "dev.env"
       - "creds.env"
-    ports:
-      - "6379:6379"
 volumes:
   postgres_data: {}


itdependsnetworks:
I like having sql server port exposed in dev, I often connect directly to the DB. Redis, I have not needed it yet.

glennmatthews:
In Nautobot core we have the ability for `invoke cli` to optionally connect to any of the containers:

```python
@task(help={"container": "Name of the container to shell into"})
def cli(context, container="nautobot"):
    """Launch a bash shell inside the running Nautobot container."""
    docker_compose(context, f"exec {container} bash", pty=True)
```

so if I need to troubleshoot postgres, I can do:

```
% invoke cli -c postgres
Running docker-compose command "exec postgres bash"
root@f4cd2d6e4f80:/# psql -U nautobot
psql (13.2 (Debian 13.2-1.pgdg100+1))
Type "help" for help.

nautobot=#
```

Would that meet your needs @itdependsnetworks?

glennmatthews (approved):

smk4664:
So, `invoke cli -c postgres` doesn't work as the psql client is not installed in the Nautobot container that chatops is using.  I don't think that should block this PR though as that is likely an upstream issue.

smk4664:
```
smk4664@stephens-mbp nautobot-plugin-chatops % invoke cli -c postgres
Running docker-compose command "ps --services --filter status=running"
Running docker-compose command "exec nautobot bash"
root@af8059a93513:/source# postgres
bash: postgres: command not found
root@af8059a93513:/source#
```

PR #86, opened 2021-08-19
Updates with service restart and post_upgrade

jvanderaa:
Adds in the missing part that had restarting of services and running post_upgrade

### DIFF
--- a/docs/chat_setup/chat_setup.md
+++ b/docs/chat_setup/chat_setup.md
@@ -63,7 +63,7 @@ TODO: clarify the above statement, starting from "The plugin behavior can be con
 
 ### Run Migrations
 
-Once the setup is done, the Nautobot `post_install` command needs to be run. As the _Nautobot user_ (which following Nautobot install docs is `nautobot`) execute:
+Once the setup is done, the Nautobot `post_upgrade` command needs to be run. As the _Nautobot user_ (which following Nautobot install docs is `nautobot`) execute:
 
 ```bash
 nautobot-server post_upgrade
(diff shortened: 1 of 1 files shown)


docs/chat_setup/chat_setup.md
```diff
@@ -61,6 +61,24 @@ The plugin behavior can be controlled with the following list of general setting
 
 TODO: clarify the above statement, starting from "The plugin behavior can be controlled . . ."
 
+### Run Migrations
+
+Once the setup is done, the Nautobot `post_install` command needs to be run. As the _Nautobot user_ (which following Nautobot install docs is `nautobot`) execute:
```
smk4664, line 66: I would suggest changing "the Nautobot `post_install` command" to "the Nautobot `post_upgrade` command".

docs/chat_setup/chat_setup.md
```diff
@@ -61,6 +61,24 @@ The plugin behavior can be controlled with the following list of general setting
...
+```
+
+This command runs migrations and clears the cache as necessary.
+
+### Restart Nautobot
+
+As a root user, restart the Nautobot services.
+
+```bash
+systemctl restart nautobot nautobot-worker
```
smk4664, line 79: If they are using nautobot-plugin-chatops <1.5.0, and have the legacy nautobot-rq-worker, they will need to restart that as well.

jvanderaa: That is installed with `nautobot-worker` as well. So that should not be a problem. If the user has migrated to using Celery now, thanks to your PR, that should not be required.

jvanderaa: But the rq-worker would need a restart for the updated code. Thinking how to handle this. I don't want to have the docs leave `nautobot-rq-worker` out there forever type thing.

smk4664: Well, I believe it will be needed for a little while, to help with transition.

jvanderaa: It's to your point all together. This is a valid part of it @smk4664 ! :)

smk4664:
I can't officially approve, but thanks for updating to also support the `nautobot-rq-worker`!

Issue #87, opened 2021-08-23
Expose a session Chatops context to improve user experience


chadell:
### Proposed Functionality

From the Chatops point of view, that could server multiple IM integrations, it would be interesting to offer a `context` that could be used to provide per user helper information. The `dispatcher` one it's dependent on the IM platform and this one should be shared across multiple IM integrations.

For instance, defining a specific time reference to run all the commands against.

We have used it in one Chatops implementation with this simple code:

```python
"""Functions for caching per-user context."""
import hashlib
from django.core.cache import cache
from . import MyChatops

def _get_cache_key(user: str) -> str:
    """Key generator for the cache, adding the plugin prefix name."""
    key_string = "-".join([MyChatops.name, user])
    return hashlib.md5(key_string.encode("utf-8")).hexdigest()  # nosec

def get_context(user: str) -> dict:
    """Return context stored for user."""
    return cache.get(_get_cache_key(user)) or {}

def set_context(user: str, updated_context: dict) -> dict:
    """Update user context."""
    context = get_context(user)
    cache.set(_get_cache_key(user), {**context, **updated_context}, timeout=86400)
```

### Use Case

Offering a Chatops `context` would facilitate that each user would customize his experience against multiple IM, and would simplify a lot of commands using this context, because parameters can be reused.

Another example could be when you run debugging against one device, setting the device first, and then just run commands that will now via the context which is the target device.

Obviously, we should expose commands to get and set this context as needed.

jvanderaa:
@chadell is this similar to #43? I agree that this would be good to include.

jvanderaa:
I'm not sure that @smk4664 has been able to progress on #43 if that is the case.

chadell:
I don't think so. This feature request is not dependent on being able to interact with multiple environments but to to offer a chatops context to the commands (that eventually can interact with multiple environments).
The `dispatcher` has a [context](https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops/dispatchers/base.py#L37), where for instance the user_id is tracked , but not sure if we should extend this context with some cached data to be consumed by the plugin commands (with some methods such as `get_context` or `set_context`, or to provide another way.
Maybe, the option of extending the `dispatcher` with an extra `context` field would do the job... the point is to be able to expose the methods to update this context.
I would raise a PR to show how this could look like

PR #88, opened 2021-08-23
Add 'session' data for Chatops to keep interesting state

chadell:
Addressing feature #87 

Simply add a `session` data per user in the `dispatcher`, so when working on the commands they can `get`, `set` or `update` the user session data and will simplify some commands.

For instance, if all commands require a `snapshot_id` to run the commands against, it's much simpler that the user defines it with a command such as `set-snapshot` that will `set` or `update` the session info, and the commands will just use this session data.

This data is not persistent, is has a max of 1 day lifetime.

### DIFF
--- a/nautobot_chatops/dispatchers/base.py
+++ b/nautobot_chatops/dispatchers/base.py
@@ -9,8 +9,6 @@ from texttable import Texttable
 
 logger = logging.getLogger("rq.worker")
 
-DEFAULT_SESSION_CACHE_TIMEOUT = 86400
-
 
 class Dispatcher:
     """Abstract base class for all chat-app dispatchers."""
--- a/nautobot_chatops/dispatchers/base.py
+++ b/nautobot_chatops/dispatchers/base.py
@@ -1,6 +1,5 @@
 """Generic base class modeling the API for sending messages to a generic chat platform."""
 import logging
-import hashlib
 from django.templatetags.static import static
 from django.core.cache import cache
 from django.conf import settings
@@ -41,8 +40,7 @@ class Dispatcher:
     def _get_cache_key(self) -> str:
         """Key generator for the cache, adding the plugin prefix name."""
         # Using __file__ as a key customization within the cache
-        key_string = "-".join([__file__, self.context.get("user_name")])
-        return hashlib.md5(key_string.encode("utf-8")).hexdigest()  # nosec
+        return "-".join([__file__, self.context.get("user_id", "generic")])
 
     @property
     def session(self) -> dict:
--- a/nautobot_chatops/dispatchers/base.py
+++ b/nautobot_chatops/dispatchers/base.py
@@ -42,31 +42,46 @@ class Dispatcher:
         # Using __file__ as a key customization within the cache
         return "-".join([__file__, self.context.get("user_id", "generic")])
 
-    @property
-    def session(self) -> dict:
-        """Return the session data stored for the user, that should be a dict."""
-        return cache.get(self._get_cache_key()) or {}
+    def get_session(self, key: str = None):
+        """Return the session data for a key, if specified, or the whole session data."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        if key:
+            # This can raise a KeyError if the key is not present
+            return session_value[key]
+        return session_value
 
-    @session.setter
-    def session(self, value: dict):
-        """Set the session data for a user."""
+    def set_session(self, value, key: str = None):
+        """Set the session data for a key, if specified, or the whole session data.
+
+        'value' can be the whole session data (dict) or the value of a specific key.
+        """
+        session_value = cache.get(self._get_cache_key()) or {}
+        if key:
+            session_value[key] = value
+        else:
+            if not isinstance(value, dict):
+                raise ValueError("Value must be a dict for the whole session data.")
+            session_value = value
+
         cache.set(
             self._get_cache_key(),
-            value,
+            session_value,
             timeout=settings.PLUGINS_CONFIG["nautobot_chatops"].get(
                 "session_cache_timeout", DEFAULT_SESSION_CACHE_TIMEOUT
             ),
         )
 
-    def update_session(self, updated_session: dict):
-        """Update the session for a user."""
-        cache.set(
-            self._get_cache_key(),
-            {**self.session, **updated_session},
-            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"].get(
-                "session_cache_timeout", DEFAULT_SESSION_CACHE_TIMEOUT
-            ),
-        )
+    def unset_session(self, key: str = None):
+        """Unset a session data for a key or as a whole."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        if key:
+            try:
+                del session_value[key]
+                self.set_session(session_value)
+            except KeyError:
+                pass
+        else:
+            cache.delete(self._get_cache_key())
 
     @classmethod
     def subclasses(cls):
--- a/nautobot_chatops/dispatchers/base.py
+++ b/nautobot_chatops/dispatchers/base.py
@@ -1,9 +1,8 @@
 """Generic base class modeling the API for sending messages to a generic chat platform."""
-
 import logging
-
+import hashlib
 from django.templatetags.static import static
-
+from django.core.cache import cache
 from django.conf import settings
 
 from texttable import Texttable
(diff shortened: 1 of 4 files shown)


nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +36,26 @@ def __init__(self, context=None):
         """Init this Dispatcher with the provided dict of contextual information (which will vary by app)."""
         self.context = context or {}
 
+    def _get_cache_key(self) -> str:
+        """Key generator for the cache, adding the plugin prefix name."""
+        # Using __file__ as a key customization within the cache
+        key_string = "-".join([__file__, self.context["user_name"]])
```
chadell, line 42:
```suggestion
        key_string = "-".join([__file__, self.context.get("user_name")])
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +36,26 @@ def __init__(self, context=None):
...
+
+    @property
+    def session(self) -> dict:
+        """Return the session data stored for the user, that should be a dict."""
+        return cache.get(self._get_cache_key()) or {}
+
+    @session.setter
+    def session(self, value: dict):
+        """Set the session data for a user."""
+        cache.set(self._get_cache_key(), value, timeout=86400)
```
jvanderaa, line 53: Timeout should be configurable with a default.

jvanderaa (changes requested):

nautobot_chatops/dispatchers/base.py
chadell: totally

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,38 @@ def __init__(self, context=None):
...
         self.context = context or {}
 
+    def _get_cache_key(self) -> str:
+        """Key generator for the cache, adding the plugin prefix name."""
+        # Using __file__ as a key customization within the cache
+        key_string = "-".join([__file__, self.context.get("user_name")])
+        return hashlib.md5(key_string.encode("utf-8")).hexdigest()  # nosec
+
+    @property
+    def session(self) -> dict:
```
glennmatthews, line 48: Instead of a single `session` dictionary, with APIs to get/set/update the dict as a whole, did you consider exposing it more directly as a key-value store? Something like:

```python
def get_session(key) -> value:
def set_session(key, value):
def unset_session(key):
```

It *feels* to me like that would be more generally useful, but you have the more direct experience with how this would be actually used, so what do you think?

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,38 @@ def __init__(self, context=None):
         """Init this Dispatcher with the provided dict of contextual information (which will vary by app)."""
         self.context = context or {}
 
+    def _get_cache_key(self) -> str:
+        """Key generator for the cache, adding the plugin prefix name."""
+        # Using __file__ as a key customization within the cache
+        key_string = "-".join([__file__, self.context.get("user_name")])
```
glennmatthews, line 44: Might be safer to use `user_id` rather than `user_name` here as I know the former should be unique, but the latter might not be guaranteed to be.

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,38 @@ def __init__(self, context=None):
...
+        return hashlib.md5(key_string.encode("utf-8")).hexdigest()  # nosec
```
glennmatthews, line 45: Is this intended as a security measure of sorts? Or else, why not just use the key_string directly as the key?

chadell: it is not needed, I used this formula to keep a consistent format of the keys, but it's not needed in this case

nautobot_chatops/dispatchers/base.py
chadell: it's an extra feature that, as you pointed out, would simplify its usage!

chadell: @glennmatthews , check this implementation https://github.com/nautobot/nautobot-plugin-chatops/pull/88/commits/c46b4cf694dbca0c7fe11a99461a4f71aedb4955
I took your idea and extended/complicated, supporting both modes, depending if a `key` is provided. I can see it useful to set/get the session data in one shoot, but I also see that this makes the methods a bit more obscure to the user.
It's kind of a tradeoff...

chadell: @glennmatthews just a heads up for this PR, to move it forward

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +37,52 @@ def __init__(self, context=None):
...
+
+    def get_session(self, key: str = None):
+        """Return the session data for a key, if specified, or the whole session data."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        if key:
+            # This can raise a KeyError if the key is not present
+            return session_value[key]
+        return session_value
+
+    def set_session(self, value, key: str = None):
```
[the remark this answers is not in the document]
chadell: Hopefully 1b2cf05 improves its usability.

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+        # Using __file__ as a key customization within the cache
+        return "-".join([__file__, self.context.get("user_id", "generic")])
+
+    def get_session_entry(self, key: str):
+        """Return the session data for a key."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        try:
+            return session_value[key]
+        except KeyError:
+            return None
```
glennmatthews, line 52: Looks like the `cache` object works basically the same as a dict, so this could be a one-liner:

```suggestion
        return cache.get(self._get_cache_key(), {}).get(key, None)
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+        session_value = cache.get(self._get_cache_key()) or {}
+        try:
+            del session_value[key]
+            self.set_session(session_value)
+        except KeyError:
+            pass
+
+    def get_session(self):
+        """Return the whole session data."""
+        return cache.get(self._get_cache_key()) or {}
```
glennmatthews, line 78:
```suggestion
        return cache.get(self._get_cache_key(), {})
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+        """Return the session data for a key."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        try:
+            return session_value[key]
+        except KeyError:
+            return None
+
+    def set_session_entry(self, key: str, value):
+        """Set the session data for a key."""
+        session_value = cache.get(self._get_cache_key()) or {}
```
glennmatthews, line 56:
```suggestion
        session_value = cache.get(self._get_cache_key(), {})
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+            self._get_cache_key(),
+            session_value,
+            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"].get(
+                "session_cache_timeout", DEFAULT_SESSION_CACHE_TIMEOUT
+            ),
+        )
+
+    def unset_session_entry(self, key: str):
+        """Unset a session data for a key."""
+        session_value = cache.get(self._get_cache_key()) or {}
```
glennmatthews, line 69:
```suggestion
        session_value = cache.get(self._get_cache_key(), {})
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -1,15 +1,16 @@
...
 from django.templatetags.static import static
-
+from django.core.cache import cache
 from django.conf import settings
 
 from texttable import Texttable
 
 logger = logging.getLogger("rq.worker")
 
+DEFAULT_SESSION_CACHE_TIMEOUT = 86400
```
glennmatthews, line 12: Let's set this value as a `default_settings` value on the `NautobotChatOpsConfig` class, as that's more consistent with the plugin conventions.

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+        """Set the session data for a key."""
+        session_value = cache.get(self._get_cache_key()) or {}
+        session_value[key] = value
+
+        cache.set(
+            self._get_cache_key(),
+            session_value,
+            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"].get(
+                "session_cache_timeout", DEFAULT_SESSION_CACHE_TIMEOUT
+            ),
```
glennmatthews, line 64: If you set the default on the `NautobotChatOpsConfig`, this can be simply:

```suggestion
            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"]["session_cache_timeout"],
```

nautobot_chatops/dispatchers/base.py
```diff
@@ -37,6 +38,63 @@ def __init__(self, context=None):
...
+        if not isinstance(value, dict):
+            raise ValueError("Value must be a dict for the whole session data.")
+        session_value = value
+
+        cache.set(
+            self._get_cache_key(),
+            session_value,
+            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"].get(
+                "session_cache_timeout", DEFAULT_SESSION_CACHE_TIMEOUT
+            ),
```
glennmatthews, line 91:
```suggestion
            timeout=settings.PLUGINS_CONFIG["nautobot_chatops"]["session_cache_timeout"],
```

glennmatthews (commented):
Nearly there, thanks!

nautobot_chatops/dispatchers/base.py
glennmatthews: Can remove this from here now as it's unused.

glennmatthews (approved):

jvanderaa (approved):

PR #90, opened 2021-09-01
Implement the CI pipeline based on GH actions

ubajze:
I have implemented the first CI pipeline based on GH actions. This pipeline follows the instructions in this repo:

https://github.com/nautobot/nornir-nautobot/blob/develop/.github/workflows/ci.yml

The pipeline was tested and publishing works.

@jvanderaa @jdrew82 @chadell I would appreciate it if you can review the configuration.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: black"
+        run: "poetry run invoke black"
```
jvanderaa, line 24:
```suggestion
        run: "invoke black"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: bandit"
+        run: "poetry run invoke bandit"
```
jvanderaa, line 41:
```suggestion
        run: "invoke bandit"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: pydocstyle"
+        run: "poetry run invoke pydocstyle"
```
jvanderaa, line 60:
```suggestion
        run: "invoke pydocstyle"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: flake8"
+        run: "poetry run invoke flake8"
```
jvanderaa, line 79:
```suggestion
        run: "invoke flake8"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: yamllint"
+        run: "poetry run invoke yamllint"
```
jvanderaa, line 98:
```suggestion
        run: "invoke yamllint"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+        uses: "actions/checkout@v2"
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
+      - name: "Linting: yamllint"
```
jvanderaa, line 97: We should have the `tasks.py` file updated to have a different format for the output of yamllint. Unless we can find out a setting within GH Actions to better output the display from the text. If you have an error on YAMLLINT you are not able to see what file has the error.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Linting"
+        run: "pip install invoke poetry"
+      - name: "Copy credentials"
+        run: "cp development/creds.example.env development/creds.env"
+      - name: "Build Container"
+        run: "poetry run invoke build"
+      - name: "Linting: Pylint"
+        run: "poetry run invoke pylint"
```
jvanderaa, line 137:
```suggestion
        run: "invoke pylint"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Linting"
+        run: "pip install invoke poetry"
+      - name: "Copy credentials"
+        run: "cp development/creds.example.env development/creds.env"
+      - name: "Build Container"
+        run: "poetry run invoke build"
+      - name: "Run Tests"
+        run: "poetry run invoke unittest"
```
jvanderaa, line 164:
```suggestion
        run: "invoke unittest"
```

jvanderaa (changes requested):
Let's have the `poetry run` part of execution removed. Since `poetry install` is run, the need for `poetry run` is not there. I could be persuaded to remove the `poetry install` step if that is a common practice to run. My personal workflow is to run poetry install before running anything. But I may not be doing it right.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Check out repository code"
+        uses: "actions/checkout@v2"
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke poetry"
+      - name: "Install Python Packages for Linting"
+        run: "poetry install"
```
chadell, line 22: As commented, it would be good to optimize it and do not install the full package, only the `invoke` and `black`.

chadell:
``` 
      - name: "Install Python Packages for Linting"
        run: "pip install --upgrade pip wheel invoke toml flake8"
      - name: "Linting: flake8"
        run: "invoke flake8"
```

.github/workflows/ci.yml
chadell: but is this a problem GH Actions is introducing? or something not related to this PR?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      fail-fast: true
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        nautobot-version: ["1.0.3", "1.1.2"]
+    runs-on: "ubuntu-20.04"
+    env:
+      INVOKE_NAUTOBOT_CHATOPS_PYTHON_VER: "${{ matrix.python-version }}"
+      INVOKE_NAUTOBOT_CHATOPS_NAUTOBOT_VER: "${{ matrix.nautobot-version }}"
+    steps:
+      - name: "Check out repository code"
```
chadell, line 151: would make sense to use YAML anchor for this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Linting"
+        run: "pip install invoke poetry"
+      - name: "Copy credentials"
+        run: "cp development/creds.example.env development/creds.env"
+      - name: "Build Container"
+        run: "poetry run invoke build"
```
chadell, line 162:
```suggestion
        run: "nvoke build"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages"
+        run: "pip install invoke poetry"
+      - name: "Set env"
+        run: "echo RELEASE_VERSION=${GITHUB_REF:10} >> $GITHUB_ENV"
+      - name: "Run Poetry Version"
+        run: "poetry version $RELEASE_VERSION"
+      - name: "Run Poetry Build"
```
chadell, line 184: eventually, we could save this build reusing the build from `build` step (not needed in this PR, but just add a comment to remember)

chadell (approved):
LGTM!

ubajze:
@jvanderaa I have replaced `poetry` with pip in the first lint jobs. I figured out that it is much faster, using pip and install just the correct lint package instead of having a full environment installation. For example, if I use `pip` to install `black` it takes 10-15 seconds to complete a job, while it takes 1min 45s if I install full environment with the `poetry`.

I think that here, were are testing our files, and it is not 100% necessary to have the correct version of `black`, `yamllint`, etc... 

Let me know what do you think...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,211 @@
...
+      INVOKE_NAUTOBOT_CHATOPS_LOCAL: "True"
+    steps:
+      - name: "Check out repository code"
+        uses: "actions/checkout@v2"
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - name: "Install Python Packages for Setup"
+        run: "pip install invoke black"
```
jvanderaa, line 20: This should be from the poetry installed version. The poetry lock file maintains the version of Black for the project. Black itself updates somewhat regularly and then will cause lots of confusion.

jvanderaa (changes requested):

.github/workflows/ci.yml
chadell: ok, then maybe we could just install the Poetry dev dependencies with `--dev-only`, so the installation process wil lbe lighter.

ubajze: Do you mean this just for `black` or other tools as well (`yamllint`, `bandit`, `pydocstyle`)? The reason I am asking this is that using `pip` saves a lot of time... 

On the other hand, do we care if we use the version `21.8b0` or `21.4b0` for example?

I also checked the cookiecutter and we have `black = "*"` in the `pyproject.toml`. So I guess we are using a different versions of `black` for different projects already (based on the time when the `poetry.lock` is created).

ubajze: > --dev-on

@chadell  I didn't see your message before. But `--dev-only` will be available in poetry `1.2`.

ubajze: > ok, then maybe we could just install the Poetry dev dependencies with `--dev-only`, so the installation process wil lbe lighter.

@chadell I didn't see your message before. But `--dev-only` will be available in poetry `1.2`.

.github/workflows/ci.yml
ubajze: @jvanderaa @chadell please check the following link:
https://github.com/nautobot/nautobot-plugin-chatops/runs/3500131352?check_suite_focus=true

Having just `invoke black` after `poetry install` is not enough. You have to start a shell or use `poetry run xxxx`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,283 @@
...
+      - name: "Install Poetry"
+        run: "pip install poetry"
+      - name: "Cache"
+        uses: "actions/cache@v2"
+        id: "cached-poetry"
+        with:
+          path: "~/.cache/pypoetry/virtualenvs/"
+          key: "{{ runner.os }}-poetry-${{ hashFiles('./poetry.lock') }}"
+      - name: "Install environment"
+        if: "steps.cached-poetry.outputs.cache-hit != 'true'"
```
chadell, line 28: 🚀

chadell (approved):
Good one with the `cache`!
I have one question to clarify. My understanding is that after the `cache`, all the data created is stored in the `path` folder. But, then, how `poetry run` knows where to find the virtualenv? (I guess I'm missing something)
On the other hand, maybe a bit picky, but I would propose 2 small changes (not dependent between them):
* Move the caching to a new step "Poetry install" to not mix with `Black`? I believe it could make it clearer
* Now I believe that makes sense to use YAML anchor, and consolidate lines 89-105 to only one. Actually, we could define the anchor in the proposed extra step above (&poetry-install) and use the anchor (<<poetry-install) in all the other steps. Ping me if you want to pair on it

.github/workflows/ci.yml
ubajze: @chadell the `cache` action first checks if the cache with the `key` exists. If not, the output variable `cache-hit` is empty. If the cache with the `key` exists then the cache is copied to the `path`. So the `path` is used only when the `key` exists. In this case the `cache-hit` is set to `true`. I know, complicated... :) I spent quite a lot of time getting the understanding. 

The `key` is generated with the variables as you can see. The key differentiator is the hash that is calculated from the `poetry.lock`. When this file is changed, the key is different and the cache does not exist. So, in that case, the new environment is installed. If the job is executed successfully, the cache is stored as the last step in the job.

We need this step in every job. At first, I thought I can have something like `setup` as the first job, but then I realized that this does not have much sense, as I will still need the `cache` action in every subsequent job.

As far as I understand, YAML anchors are not supported in GitHub actions. But, you can create the `composite` action and then reuse that in your jobs. Can you please check the `gha-custom` branch. I tried to implement this logic. So, I have the action that setup the environment -> either load the environment from the cache or install the environment using `poetry install`.

ubajze:
@jvanderaa I migrated all jobs to `poetry install`, but I am using caching. So, the execution time is similar to `pip install`. Can you please check, and let me know what do you think.

.github/workflows/ci.yml
chadell: I like the `gha-custom` branch! is there any issue to adopt it?

.github/workflows/ci.yml
jvanderaa: That mechanism as shown is not going to work @chadell , that will install the latest vs what is defined in the dev install. I think we have a working pattern though.

jvanderaa:
LGTM. Merging in.

jvanderaa (approved):

PR #92, opened 2021-09-06
Custom GH action to replace repeatable steps

ubajze:
This PR includes the custom GH action, which replaces repeatable steps in every job.

The custom action is called `setup-environment`. The action checks if the cache exists for the particular lock file. If that is true, then the environment is loaded from the cache otherwise, the fresh environment is installed.


chadell (approved):
much more readable!

jvanderaa (changes requested):
Looks good.

In double checking one of the requests from the last PR I realized that we are missing the update to Yamllint that was requested. Can you make that as part of this PR? 

Within `def yamllint`, make the command: 

```
yamllint . --format standard
```

jvanderaa (approved):

PR #91, opened 2021-09-02
Two new chat commands

tim-fiola:
This PR has 2 new chat sub-commands:

`get-manufacturer-summary` - gives a summary of each manufacturer and how many devices have that manufacturer

```
Manufacturer   Quantity of Devices
==================================
arista         5                  
cisco          5                  
hp             0                  
juniper        3                  
mellanox       0                  
meraki         0                  
nvidia         0                  
opengear       0                  
pan            0       
```

`get-circuit-connections` - tells the user what Devices/Circuits a given Circuit connects to.  

image

PR #98, opened 2021-09-14
Two new commands

tim-fiola:
Closed #91 to this request, hopefully to clear up some weird GitHub rendering diffs.

This PR has 2 new chat sub-commands:

`get-manufacturer-summary` - gives a summary of each manufacturer and how many devices have that manufacturer

```
Manufacturer   Quantity of Devices
==================================
arista         5                  
cisco          5                  
hp             0                  
juniper        3                  
mellanox       0                  
meraki         0                  
nvidia         0                  
opengear       0                  
pan            0       
get-circuit-connections - tells the user what Devices/Circuits a given Circuit connects to.
```

`get-circuit-connections` - returns what a circuit connects to on each end (exclusive of any cables)

image


nautobot_chatops/workers/nautobot.py
```diff
@@ -999,3 +1025,150 @@ def about(dispatcher, *args):
...
+            [("Provider Name", provider.slug), ("Circuit ID", circuit.cid)],
+            "circuit connection info",
+            nautobot_logo(dispatcher),
+        )
+    )
+
+    header = ["Side", "Connecting Object"]
+    rows = [("A", endpoint_info_a), ("Z", endpoint_info_z)]
+
+    dispatcher.send_large_table(header, rows)
```
glennmatthews, line 1152: Given that there are only two rows here, use of send_large_table is a bit overkill, and I think not the most visually appealing way to present this information. Perhaps consider either:
1) Adding these directly to the command_response_header (they're not technically "args", but they are key-value pairs that you could append after the Provider Name and Circuit ID, and it might be more attractive that way)
2) Just sending these directly as a block of markdown text (`f"Side A: {endpoint_info_a}"`, `f"Side Z: {endpoint_info_z}"`)

nautobot_chatops/workers/nautobot.py
```diff
@@ -999,3 +1025,150 @@ def about(dispatcher, *args):
...
+        dispatcher.send_error(cid_not_found_msg)
+        return (CommandStatusChoices.STATUS_FAILED, cid_not_found_msg)
+
+    if menu_item_check(circuit_id):
+        prompt_for_circuit(
+            "nautobot get-circuit-connections",
+            "Get Nautobot Circuit Connections",
+            dispatcher,
+            offset=menu_offset_value(circuit_id),
+        )
```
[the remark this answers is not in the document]
tim-fiola: OOPS.  Yeah, let me take a look at that.  I did a code refactor and this may have gotten goofed.

nautobot_chatops/workers/nautobot.py
tim-fiola: Hi Glenn.  I tried both your suggestions.
1. Adding these directly to the command_response_header gives a bad shortcut output
image
2. Doing the straight markdown was OK, but I think it was harder to read.  I actually prefer to see the results in a table (even though it's not a large table) because of its column format

nautobot_chatops/workers/nautobot.py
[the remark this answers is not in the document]
tim-fiola: OK, fixed it.  Thank you for that catch!

nautobot_chatops/workers/nautobot.py
glennmatthews: Understood. Thanks for indulging me!

nautobot_chatops/workers/nautobot.py
```diff
@@ -62,6 +62,20 @@ def prompt_for_vlan(action_id, help_text, dispatcher, filter_type, filter_value_
...
+        return (CommandStatusChoices.STATUS_FAILED, "No Circuit Providers were found")
+    if circuits is None:  # A circuit was not provided; provide list of circuits for user to choose from
+        circuits = Circuit.objects.filter(provider__name=provider)
+    if not circuits:  # There were no Circuit objects found; notify user
+        dispatcher.send_error("No circuits were found")
+        return (CommandStatusChoices.STATUS_FAILED, "No circuits were found")
+    choices = [(f"{circuit.provider.slug}: {circuit.cid}", circuit.cid) for circuit in circuits]
+    return dispatcher.prompt_from_menu(action_id, help_text, choices, offset=offset)
+
+
```
glennmatthews, line 78: Is this entire function unused now? I don't see where it's being called.

nautobot_chatops/workers/nautobot.py
```diff
@@ -133,6 +147,39 @@ def get_filtered_connections(device, interface_ct):
     )
 
 
+def analyze_circuit_endpoints(endpoint):
+    """Analyzes a circuit's endpoint and returns info about what object the endpoint connects to."""
+    if type(endpoint) in [Interface, FrontPort, RearPort]:
+        # Put into format: object.device_name
+        info = f"Device: {endpoint.device.name}  Interface: {endpoint.name}"
```
glennmatthews, line 154: Is it desirable to refer to FrontPort and RearPort as Interfaces? Or might it be better to do something like:

```suggestion
        info = f"Device: {endpoint.device.name}  {endpoint.__class__.__name__}: {endpoint.name}"
```

nautobot_chatops/workers/nautobot.py
```diff
@@ -133,6 +147,39 @@ def get_filtered_connections(device, interface_ct):
     )
 
 
+def analyze_circuit_endpoints(endpoint):
+    """Analyzes a circuit's endpoint and returns info about what object the endpoint connects to."""
+    if type(endpoint) in [Interface, FrontPort, RearPort]:
```
glennmatthews, line 152: I believe this would be more idiomatic:

```suggestion
    if isinstance(endpoint, (Interface, FrontPort, RearPort)):
```

nautobot_chatops/workers/nautobot.py
```diff
@@ -133,6 +147,39 @@ def get_filtered_connections(device, interface_ct):
...
+    except (AttributeError, IndexError):
+        term_z = "No Z Side Connection in Database"
+    if term_a != "No A Side Connection in Database":
+        endpoint_info_a = analyze_circuit_endpoints(term_a)
+    else:
+        endpoint_info_a = term_a
+    if term_z != "No Z Side Connection in Database":
+        endpoint_info_z = analyze_circuit_endpoints(term_z)
+    else:
+        endpoint_info_z = term_z
```
glennmatthews, line 179: Perhaps:
```suggestion
    try:
        term_a = circuit.termination_a.trace()[0][2]
        endpoint_info_a = analyze_circuit_endpoints(term_a)
    except (AttributeError, IndexError):
        endpoint_info_a = term_a = "No A Side Connection in Database"
    try:
        term_z = circuit.termination_z.trace()[0][2]
        endpoint_info_z = analyze_circuit_endpoints(term_z)
    except (AttributeError, IndexError):
        endpoint_info_z = term_z = "No Z Side Connection in Database"
```

nautobot_chatops/workers/nautobot.py
tim-fiola: made this change.  Thank you!

nautobot_chatops/workers/nautobot.py
tim-fiola: Thanks for pointing that out.   I condensed/simplified the code

glennmatthews (approved):
Nice work!

tim-fiola:
> Nice work!

Thanks for your reviews, Glenn!  Super helpful.

jvanderaa (approved):

PR #101, opened 2021-09-17
Migrate to the action in its own repo

ubajze:
Migration to this action:

https://github.com/networktocode/gh-action-setup-poetry-environment


glennmatthews (approved):
Looks reasonable to me

Issue #106, opened 2021-09-28
Implement maximum message size exceeded check for MS Teams.


progala:
### Environment
* Nautobot version:  1.0.2+
* nautobot-plugin-chatops version:  1.5.1

### Proposed Functionality

Proposal is to add a new setting "MICROSOFT_MAX_MSG_SIZE" that would be used to check if the prepared content exceeds the provided value.

If the size of the prepared content exceeds the specified value, then the error message would be sent to MS Teams chat and an warning would be logged internally.

### Use Case

When number of objects is large response generated by some of the chatbot commands exceed maximum allowed message size in MS Teams.

The problem is that Nautobot emits a message but the message never makes it to MS Teams. Ocasionally part of the response, like reponse header, might be received, with the rest of the message not showing up in the chat. This represents a silent failure that makes it difficult to debug and reason about, i.e. it's not apparent to the end user what, if anything, is wrong.

PR #107, opened 2021-09-28
Implement maximum msg size check for MS Teams.

progala:
Draft proposal of changes needed to implement #106 .

- Add a new setting `microsoft_max_msg_size` which controls maximum allowed message size for MS Teams.
- Set value of `microsoft_max_msg_size` suing `MICROSOFT_MAX_MSG_SIZE` env var. Default to 0.
- Modify MS Teams dispatcher to 
  - perform `json.dumps()` on content before message is sent
  - check the size of resulting string does not exceed `microsoft_max_msg_size` value
  - if the size is too large, return error message to MS Teams chat, and log a warning
  - otherwise continue as normal

### DIFF
--- a/development/nautobot_config.py
+++ b/development/nautobot_config.py
@@ -217,6 +217,7 @@ PLUGINS_CONFIG = {
         "enable_webex": True,
         "microsoft_app_id": os.environ.get("MICROSOFT_APP_ID"),
         "microsoft_app_password": os.environ.get("MICROSOFT_APP_PASSWORD"),
+        "microsoft_max_msg_size": int(os.environ.get("MICROSOFT_MAX_MSG_SIZE", 0)),
         "slack_api_token": os.environ.get("SLACK_API_TOKEN"),
         "slack_signing_secret": os.environ.get("SLACK_SIGNING_SECRET"),
         "slack_slash_command_prefix": os.environ.get("SLACK_SLASH_COMMAND_PREFIX", "/"),
--- a/nautobot_chatops/dispatchers/ms_teams.py
+++ b/nautobot_chatops/dispatchers/ms_teams.py
@@ -1,6 +1,7 @@
 """Dispatcher implementation for sending content to Microsoft Teams."""
 import os
 import logging
+import json
 import requests
 
 from django.conf import settings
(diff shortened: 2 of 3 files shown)


development/nautobot_config.py
```diff
@@ -217,6 +217,7 @@ def is_truthy(arg):
         "enable_webex": True,
         "microsoft_app_id": os.environ.get("MICROSOFT_APP_ID"),
         "microsoft_app_password": os.environ.get("MICROSOFT_APP_PASSWORD"),
+        "microsoft_max_msg_size": int(os.environ.get("MICROSOFT_MAX_MSG_SIZE", 0)),
```
glennmatthews, line 220: Do we know what the actual max message size is for MS Teams? Does this even need to be a configurable parameter?

nautobot_chatops/dispatchers/ms_teams.py
```diff
@@ -85,10 +86,21 @@ def _send(self, content, content_type="message"):
                 "replyToId": self.context["message_id"],
             }
         )
+        json_content = json.dumps(content)
+        max_msg_size = settings.PLUGINS_CONFIG["nautobot_chatops"]["microsoft_max_msg_size"]
+        if 0 < max_msg_size < len(json_content):
+            msg = "Prepared content, size %s, exceeded maximum allowed MS Teams msg size." % len(json_content)
+            logger.warning(msg)
+            json_content = json.dumps(
+                self.send_error(
```
glennmatthews, line 95: I don't think I like how this recurses - in `_send`, we're now calling `send_error`, which calls `send_blocks`, which calls `_send`. Won't that result in a duplicate message being sent in the best case, and infinite recursion in the worst case (where `max_msg_size < len(json.dumps(self.send_error(...)))`)?

I think it'd be better and safer to just construct the block directly here, i.e.:
```python
json_content = json.dumps([{
    "type": "TextBlock",
    "text": "Message size exceeds maximum allowed by MS Teams.\nConsider changing the filtering method.",
    "wrap": True,
    "color": "attention",
}])
```

development/nautobot_config.py
smk4664: From MS Teams: 28 KB is an approximate limit because it includes the message itself (text, image links, etc.), @-mentions, number of connectors, and reactions.

Looks like character length has nothing to do with this.  Maybe there is a message coming back to the chatops plugin from Microsoft that would be better used.

progala: I've not observed any messages coming back from Microsoft. I agree that counting character length is not ideal but in our case I believe it will the main contributor to the message size. From testing it looked to me that messages started getting lost when content length exceeded approx. 21000.

nautobot_chatops/dispatchers/ms_teams.py
progala: That's a good point, constructing the block directly would be much safer.

development/nautobot_config.py
progala: @glennmatthews @smk4664 I did some more digging and I identified some responses we could use, but they're not consistent.

E.g. when using `nautobot get-device-facts` with large number of devices I get http code 413 back with a telling message:

```
worker_1    | 19:10:38 INFO Response code 413 - rq.worker - ms_teams - /source/nautobot_chatops/dispatchers/ms_teams.py:111
worker_1    | b'{"error":{"code":"MessageSizeTooBig","message":"Message size too large."}}'
```

When I ran `nautobot get-vlans site <site-name>` for site which had a lot of vlans a message header made it to the chat but VLAN list did not and I could see in debug http code 400 with a generic error message:

```
worker_1    | 19:10:06 INFO Response code 400 - rq.worker - ms_teams - /source/nautobot_chatops/dispatchers/ms_teams.py:111
worker_1    | b'{"error":{"code":"ServiceError","message":"Unknown"}}'
```

progala:
I'll work on imlementing proper error handling based on the status codes returned by MS Teams instead of the approach taken in this PR.

PR #111, opened 2021-10-30
Updates CI to include proper on release

jvanderaa:
It was brought up that 1.5.1 release did not publish to Pypi properly. Looking through the CI setup, it is missing the releases section. This adds releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,14 @@
 ---
 name: "CI"
-on:  # yamllint disable
-  - "push"
-  - "pull_request"
+on: # yamllint disable
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


glennmatthews (commented):
I don't see this changeset in the other Nautobot plugins I've checked (ssot, circuit-maintenance, golden-config). Are they wrong too? Where did this fix come from?

jvanderaa:
It came from the Nautobot-Ansible repo. I will try to re-release the 1.5.1 branch to see if that will re-trigger properly. We only have 1.5.0 on pypi at the moment.

jvanderaa:
Looks like it is now working semi-properly. Just need to figure out the creds part.

PR #113, opened 2022-01-04
 Update Slack docs to support new manifest editor

matt852:
I've updated the docs for setting up the chatops plugin with Slack to support their new Manifest importer.

I've also included a properly sized Nautobot icon that can be used with Slack.  Unlike MS Teams, this cannot (currently) be imported with a single .zip file, so it is separate.


matt852:
I'm not sure why the workflow failed, as the `invoke black` command works just fine locally.
```bash
(nautobot-chatops-B-B2krGD-py3.9) ➜  nautobot-plugin-chatops git:(update-slack-docs) ✗ invoke black
Running docker-compose command "ps --services --filter status=running"
Running docker-compose command "run --entrypoint 'black --check --diff .' nautobot"
Creating network "nautobot-chatops_default" with the default driver
Creating nautobot-chatops_redis_1    ... done
Creating nautobot-chatops_postgres_1 ... done
Creating nautobot-chatops_nautobot_run ... done
All done! ✨ 🍰 ✨
52 files would be left unchanged.
```

smk4664:
@matt852 This is an issue with the typed-ast library which comes from Black and the Python Version. I had to do a `poetry lock` in another repo to update the typed-ast library.  Since this is docs only, we could probably ignore this for now, but it will need to be fixed.  @jvanderaa What are your thoughts on this?

docs/chat_setup/slack_setup.md
```diff
@@ -27,9 +28,29 @@ PLUGINS_CONFIG = {
...
+   - On line 12, under setting `features/slash_commands/url`, update `<your-URL>` with a publicly accessible URL to your Nautobot server. Note: The trailing `/api/plugins/...` are required.
+   - Repeat this for line 31, under setting `settings/interactivity/request_url`
+4. Review the summarized settings on the next window and click Create.
+5. On the General --> Basic Information page, note the `Signing Secret` near the bottom, under App Credentials. This will be needed later for setting `SLACK_SIGNING_SECRET`.
+6. Under Distribution --> Install App, select `Install to Workspace`.
+   - If you are not a Slack admin, this step will require approval first from an Slack admin in your company.
+   - Once installed, note the `Bot User OAuth Token` here. This will be needed later for setting `SLACK_API_TOKEN`.
+7. Continue with below section "Post App-Creation Steps"
+
+> **Optional:** You can configure the App Icon on the General --> Basic Information page. Under `App Icon`, select "Choose File." You can use the supplied icon [nautobot_chatops_icon.png](https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops_icon.png).
```
smk4664, line 46: The link here is giving me a 404.

docs/chat_setup/slack_setup.md
```diff
@@ -130,4 +158,6 @@ PLUGINS_CONFIG = {
 
 > NOTE: Custom chatbot development allows for chatbot slash commands such as `/grafana` and `/meraki`.
 
-[Chat Setup](./chat_setup.md)
+## General Chat Setup Instructions
+
+Set [Chat Setup](./chat_setup.md) instructions here for general plugin setup instructions.
```
smk4664, line 163: See [Chat Setup]

smk4664 (changes requested):
A few items that need reviewed but thanks for the PR. This is really helpful.

glennmatthews:
> @matt852 This is an issue with the typed-ast library which comes from Black and the Python Version. I had to do a `poetry lock` in another repo to update the typed-ast library. Since this is docs only, we could probably ignore this for now, but it will need to be fixed. @jvanderaa What are your thoughts on this?

You should be able to update to the latest version of Black to avoid this issue, and if necessary, update the pyproject.toml as seen in https://github.com/nautobot/nautobot/pull/1105/files

docs/chat_setup/slack_setup.md
matt852: Good catch.  Updated with the correct subdirectory.

However this directory/file is part of this PR, so will give a broken link if clicked on directly until this is merged.

docs/chat_setup/slack_setup.md
matt852: Fixed

matt852:
I fixed the CI issues with the help of @nniehoff.  I went ahead and fixed the outstanding `black` and `pylint` errors discovered as well, as a result of upgraded those packages.

pyproject.toml
```diff
@@ -24,9 +24,14 @@ texttable = "^1.6.2"
 PyJWT = "^2.1.0"
 webexteamssdk = "^1.3"
 slack-sdk = "^3.4.2"
+celery = [
```
jvanderaa, line 27: Will this work with just running an update instead of pinning versions?

matt852: No.  There are issues with 5.2.3 and setuptools, released on 12/29/2021.  See release notes [here](https://github.com/celery/celery/releases/tag/v5.2.3).

In it, they pin the version of setuptools, which breaks compatibility with Nautobot.  There are also Markdown related version issues, which is required to be set to !=3.3.5 to also prevent dependency issues.

jvanderaa (approved):
We will update this again when Nautobot has some updates. This is good to go.

Issue #115, opened 2022-01-14
Better (sub)command permission management


david-kn:
### Environment
* Nautobot version: 1.2.2
* nautobot-plugin-chatops version: 1.5.1

### Proposed Functionality
Add ability to define permissions for certain subcommands via regex / in 'groups' and not only individually.

These possible approaches come to my mind:
1. support regex expressions in (sub)commands, e.g. `^get-` would match all `get `(read-only) subcommands
2. or support multiple values in field (comma separated values?), e.g. `get-device-status, get-devices`

### Use Case
Different teams and people in different channels may need different and limited rights. It would be very helpful to be able set, e.g "read-only" commands (commands starting with `get-` only) within the whole channel; then allow other subset of commands for other channel (team) etc. The `*` (any) symbol is good for admin accounts but when we were considering a finer permission matrix, it's not that helpful as a regular regex support.

More specific example: At this moment, I want to enable only "read-only" (`get`) commands for a single channel, I would have to type 10 records (each for allowing a single get command). And more channels and more specific people (teams) with different rights would quickly increase the size of permission table so it would not be easily visible who, can do what and where.
- And if an update (new command) would appear, I'd have to manually add this command again
- When considering use of more chatops integration like `/ipfabric` , `/grafana`, then the table would be even larger to support read-only stuff per each integration per specific channel

And regex support could help to better manage read, write, or partial rights across all of these chatops integrations.

jvanderaa:
@glennmatthews would love some insight/thought on this.

glennmatthews:
I like the idea. I'll take a look at the PR.

PR #116, opened 2022-01-18
Adding regex support for (sub)commands

david-kn:
Hi, I just quickly drafted a possible solution for regex matching in (sub)commands which could address issue #115 .

I'm not that familiar with django + their queries and as well this might not be the best possible approach but the logic and sense is incorporated to kick this off ...

Any opinions or comments are welcomed :-)

### DIFF
--- a/nautobot_chatops/models.py
+++ b/nautobot_chatops/models.py
@@ -9,6 +9,26 @@ from nautobot.core.models import BaseModel
 
 from .choices import AccessGrantTypeChoices, CommandStatusChoices, CommandTokenPlatformChoices
 
+import re
+
+
+class QuerySetRegexMatch(models.QuerySet):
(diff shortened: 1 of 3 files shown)


nautobot_chatops/models.py
```diff
@@ -9,6 +9,26 @@
...
+    def match_re(self, value):
+        records = self.all()
+
+        results = []
+        for row in records:
+            if row.subcommand == "*":  # skip as it is not a valid regex expression and would raise exception
+                continue
+            if re.match(row.subcommand, value):
+                results.append(row.subcommand)
+        return results
```
glennmatthews, line 30: I'm definitely a bit concerned about potential performance impact of this design. It *should* be possible to do this as an actual DB query, rather than querying for `all()` and then iterating through the results, though from some quick Googling it looks like custom SQL might be needed to make that possible...

PR #120, opened 2022-01-21
Uses settings to control ephemeral state.

### LINKED ISSUE #114: Do not show command results to all channel members (reduce result visibility)
### Environment
* Nautobot version:  1.2.2 
* nautobot-plugin-chatops version: 1.5.1

### Proposed Functionality
Send all the command results visible only to person triggering the command (currently it is visible to the whole channel). Ideally this could be configurable in admin section / plugin configuration (in order not to the change your design concept entirely and keep the ability to choose)?
It would be useful if this would / could influence also other relevant nautobot chatops integrations (IP Fabric, Grafana etc.)

### Use Case
The channel is getting easily messy with several command outputs from different team members therefore it is hard to orient in such a channel + also people do not tend to search for older messages but rather type their own. 
Also the data that were provided, let's say yesterday, doesn't have to be still valid and person is sending command again anyway because one want to be certain about up-to-date data.

jvanderaa:
Fixes #114 

Moves to use settings for ephemeral settings across all of the dispatchers for what was previously `ephemeral=False`.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -162,4 +162,6 @@ Some known limitations of currently supported platforms:
 - No table functionality in cards (blocks). The `ColumnSet` card layout feature allows for wrapping content across
   multiple columns, but does not provide for any alignment across columns, so it's not suitable for tables.
 
+## Settings
 
+The setting of `send_all_messages_private` within the configuration applied within the Nautobot config is used to send all messages as private messages. The messages will be sent in private when this is set to `True`. The default setting is `False`, which is the behavior for several message settings.
--- a/docs/design.md
+++ b/docs/design.md
@@ -108,10 +108,8 @@ In general, we recommend structuring commands as a two-tiered command-subcommand
 every command as a top-level worker function. (`/nautobot get-device-info <device>`, `/nautobot get-vlan-info <vlan>`, etc.
 rather than `/nautobot-get-device-info <device>`, `/nautobot-get-vlan-info <vlan>`, etc.) This is because:
 
-a. On platforms such as Slack, each separate slash-command must be enabled and configured separately on the server,
-so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
-b. Platforms such as Microsoft Teams may limit the number of top-level commands that are displayed to the user in
-a chat client, so large numbers of commands may be difficult to discover.
+- On platforms such as Slack, each separate slash-command must be enabled and configured separately on the server, so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
+- Platforms such as Microsoft Teams may limit the number of top-level commands that are displayed to the user in a chat client, so large numbers of commands may be difficult to discover.
 
 That said, the implementation of Nautobot allows it to transparently support both syntaxes (`/command-sub-command` as
 well as `/command sub-command`; if the deployer takes the time to set up the bot accordingly.
@@ -171,8 +169,9 @@ The setting of `send_all_messages_private` within the configuration applied with
 
 > This table represents the platform support of particular settings
 
-| Platform | send_all_messages_private |
-| Slack | ✅ |
-| MS Teams | ❌ |
-| WebEx | ❌ |
-| Mattermost | ✅ |
+| Platform   | send_all_messages_private |
+| ---------- | ------------------------- |
+| Slack      | ✅                        |
+| MS Teams   | ❌                        |
+| WebEx      | ❌                        |
+| Mattermost | ✅                        |
--- a/docs/design.md
+++ b/docs/design.md
@@ -32,7 +32,7 @@ The design goal of this plugin is to be able to write chatbot commands once and 
 
 2. worker (`nautobot_chatops.workers`)
 
-   - This layer is *completely ignorant* of chat platforms. All code in this layer does not know or care about the
+   - This layer is _completely ignorant_ of chat platforms. All code in this layer does not know or care about the
      difference between Slack, WebEx, Microsoft Teams, or any other platform we may support in the future.
 
    - Each `job` worker function acts on the provided parameters, then invokes generic methods on its provided
@@ -109,24 +109,25 @@ every command as a top-level worker function. (`/nautobot get-device-info <devic
 rather than `/nautobot-get-device-info <device>`, `/nautobot-get-vlan-info <vlan>`, etc.) This is because:
 
 a. On platforms such as Slack, each separate slash-command must be enabled and configured separately on the server,
-   so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
+so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
 b. Platforms such as Microsoft Teams may limit the number of top-level commands that are displayed to the user in
-   a chat client, so large numbers of commands may be difficult to discover.
+a chat client, so large numbers of commands may be difficult to discover.
 
 That said, the implementation of Nautobot allows it to transparently support both syntaxes (`/command-sub-command` as
 well as `/command sub-command`; if the deployer takes the time to set up the bot accordingly.
 
 ### Multi-word Parameters
 
-Nautobot dispatchers now allow multi-word arguments to be passed into commands.  An example of this is passing city
-names to a subcommand parameters.  As an example, say we have a command that perfoms a lookup for all sites in Nautobot
-that match a city.  The command and parameters might look like `/nautobot get-sites location Dallas` where Dallas is the
-city we want to search for.  For the command to support cities such as `Las Vegas` we would want to quote the city
-argument.  The new command should look as `/nautobot get-sites location 'Las Vegas'`.  
+Nautobot dispatchers now allow multi-word arguments to be passed into commands. An example of this is passing city
+names to a subcommand parameters. As an example, say we have a command that perfoms a lookup for all sites in Nautobot
+that match a city. The command and parameters might look like `/nautobot get-sites location Dallas` where Dallas is the
+city we want to search for. For the command to support cities such as `Las Vegas` we would want to quote the city
+argument. The new command should look as `/nautobot get-sites location 'Las Vegas'`.
 
-The worker would need to preserve the quoting when prompting for additional parameters.  Below is an example:
+The worker would need to preserve the quoting when prompting for additional parameters. Below is an example:
 
 Here we use the previous example, but add limit to the site lookup.
+
 ```Python
 action = f"get-sites location '{city}'" # Adding single quotes around city to preserve quotes.
 dispatcher.prompt_for_text(action_id=action, help_text="Please enter the maximum number of sites to return.", label="Number")
@@ -157,11 +158,21 @@ Some known limitations of currently supported platforms:
 
 - No support for preformatted text in cards (blocks) - while text can be rendered as `monospace`, it still does not
   preserve whitespace, so it's not suitable for aligning output into text tables and the like.
-- While text-only messages *do* support Markdown preformatted text, the text is wrapped to a max width of 69
+- While text-only messages _do_ support Markdown preformatted text, the text is wrapped to a max width of 69
   characters regardless of how large the client's window is.
 - No table functionality in cards (blocks). The `ColumnSet` card layout feature allows for wrapping content across
   multiple columns, but does not provide for any alignment across columns, so it's not suitable for tables.
 
 ## Settings
 
-The setting of `send_all_messages_private` within the configuration applied within the Nautobot config is used to send all messages as private messages. The messages will be sent in private when this is set to `True`. The default setting is `False`, which is the behavior for several message settings.
+The setting of `send_all_messages_private` within the configuration applied within the Nautobot config is used to send all messages as private messages. The messages will be sent in private when this is set to `True`. The default setting is `False`, which is the default behavior for several message settings.
+
+### Settings - Platform Support of Settings
+
+> This table represents the platform support of particular settings
+
+| Platform | send_all_messages_private |
+| Slack | ✅ |
+| MS Teams | ❌ |
+| WebEx | ❌ |
+| Mattermost | ✅ |
--- a/nautobot_chatops/__init__.py
+++ b/nautobot_chatops/__init__.py
@@ -46,9 +46,9 @@ class NautobotChatOpsConfig(PluginConfig):
         "mattermost_api_token": None,
         "mattermost_url": None,
         # As requested on https://github.com/nautobot/nautobot-plugin-chatops/issues/114 this setting is used for
-        # sending all messages as an emphemeral message, meaning only the person interacting with the bot will see the
+        # sending all messages as an ephemeral message, meaning only the person interacting with the bot will see the
         # responses.
-        "send_all_messages_emphemeral": False,
+        "send_all_messages_private": False,
     }
 
     max_version = "1.999"
--- a/nautobot_chatops/__init__.py
+++ b/nautobot_chatops/__init__.py
@@ -45,6 +45,10 @@ class NautobotChatOpsConfig(PluginConfig):
         # Mattermost-specific settings
         "mattermost_api_token": None,
         "mattermost_url": None,
+        # As requested on https://github.com/nautobot/nautobot-plugin-chatops/issues/114 this setting is used for
+        # sending all messages as an emphemeral message, meaning only the person interacting with the bot will see the
+        # responses.
+        "send_all_messages_emphemeral": False,
     }
 
     max_version = "1.999"
--- a/nautobot_chatops/dispatchers/adaptive_cards.py
+++ b/nautobot_chatops/dispatchers/adaptive_cards.py
@@ -126,7 +126,7 @@ class AdaptiveCardsDispatcher(Dispatcher):
             blocks,
             callback_id=callback_id,
             modal=True,
-            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_emphemeral"],
+            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
             title=dialog_title,
         )
 
--- a/nautobot_chatops/dispatchers/adaptive_cards.py
+++ b/nautobot_chatops/dispatchers/adaptive_cards.py
@@ -126,7 +126,7 @@ class AdaptiveCardsDispatcher(Dispatcher):
             blocks,
             callback_id=callback_id,
             modal=True,
-            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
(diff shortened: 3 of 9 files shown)


smk4664:
I will want to test this against the chat platforms. I am concerned about some of the Mattermost commands not working in ephemeral messages.

jvanderaa:
Makes sense @smk4664. Note that I left the `ephemeral=True` settings that were already there. This was effectively a find and replace of `ephemeral=False` to being `ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_emphemeral"]`. I will also need to run through some tests, but the concept is what I was hoping to review first.

nautobot_chatops/__init__.py
```diff
@@ -45,6 +45,10 @@ class NautobotChatOpsConfig(PluginConfig):
         # Mattermost-specific settings
         "mattermost_api_token": None,
         "mattermost_url": None,
+        # As requested on https://github.com/nautobot/nautobot-plugin-chatops/issues/114 this setting is used for
+        # sending all messages as an emphemeral message, meaning only the person interacting with the bot will see the
+        # responses.
+        "send_all_messages_emphemeral": False,
```
glennmatthews, line 51: `emphemeral` -> `ephemeral` throughout -- or maybe change to `private` instead? Not sure if "ephemeral" is widely used across platforms or if it's a Slack-ism?

jvanderaa: Updated to private.

nautobot_chatops/dispatchers/base.py
```diff
@@ -216,11 +216,22 @@ def ask_permission_to_send_image(self, filename, action_id):
 
     # Send various content to the user or channel
 
-    def send_markdown(self, message, ephemeral=False):
+    def send_markdown(
+        self,
+        message,
+        ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
```
glennmatthews, line 222: Design-wise I'm not a big fan of having a function's argument defaults depend on a dynamic settings value, in particular as it makes unit testing a bit more difficult  (using Django's `override_settings` decorator probably won't work as expected here). I'd prefer a pattern more like:


```python
def send_markdown(self, message, ephemeral=None):
    if ephemeral is None:
        ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"]
   ...
```

even though it means writing a bit more code.

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -461,7 +472,13 @@ def prompt_for_text(self, action_id, help_text, label, title="Your attention ple
         }
         blocks = [textentry]
         # In Mattermost, a textentry element can ONLY be sent in a modal Interactive dialog
-        return self.send_blocks(blocks, callback_id=action_id, ephemeral=False, modal=True, title=title)
+        return self.send_blocks(
+            blocks,
+            callback_id=action_id,
+            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
```
glennmatthews, line 478: Per the Mattermost comment just above, I'd think that this one needs to stay `ephemeral=False` - but maybe I'm mistaken? Definitely would want to see this tested with Mattermost.

nautobot_chatops/dispatchers/slack.py
```diff
@@ -400,7 +409,13 @@ def multi_input_dialog(self, command, sub_command, dialog_title, dialog_list):
 
                 blocks.append(self._input_block(action_id, dialog["label"], textentry, dialog.get("optional", False)))
 
-        return self.send_blocks(blocks, callback_id=callback_id, modal=True, ephemeral=False, title=dialog_title)
+        return self.send_blocks(
+            blocks,
+            callback_id=callback_id,
+            modal=True,
+            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
```
glennmatthews, line 416: Similar concern/question here - I think a modal=True dialog is implicitly ephemeral/private in Slack, but not sure if `ephemeral=True` will actually work here or if it's an API error.

jvanderaa:
@glennmatthews are you good with the overall design (after I fix the calls design to what you have mentioned)? I think this will cover all of the bots then and I will only need to make the update here and not in each chatops plugin implementation.

glennmatthews:
I'm OK with the idea of it - just want to have confidence that the specific changes being made are tested :-)

nautobot_chatops/dispatchers/slack.py
jvanderaa: I went ahead and moved back to False all together. You are right that it should just be part of it.

nautobot_chatops/dispatchers/mattermost.py
jvanderaa: Reverted, for the same reason below.

jvanderaa:
Something that I've done in here has broken things. Working through it.

jvanderaa:
Looking at what the code is used heavily in the additional chat framework that this request is for, the Slack `send_snippet` does not support sending messages ephemerally, or directly to an individual. 

https://api.slack.com/methods/files.upload

The method `send_large_table()` uses `send_snippet` after building the table.

jvanderaa:
This PR will work for Python3.7 implementation and higher at the moment, which should also work for Python3.6. But we do not have the desired state for this yet. 

In order to accomplish the original goals, we may need to look at adding a DM feature.

nautobot_chatops/dispatchers/adaptive_cards.py
```diff
@@ -121,7 +122,13 @@ def multi_input_dialog(self, command, sub_command, dialog_title, dialog_list):
             ],
         }
         blocks.append(buttons)
-        return self.send_blocks(blocks, callback_id=callback_id, modal=True, ephemeral=False, title=dialog_title)
+        return self.send_blocks(
+            blocks,
+            callback_id=callback_id,
+            modal=True,
+            ephemeral=settings.PLUGINS_CONFIG["nautobot_chatops"]["send_all_messages_private"],
```
glennmatthews, line 129: Can an adaptive-cards multi-input dialog even be sent as ephemeral=True? I *think* dialogs are always private to a specific user, but would want to see this tested.

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -403,11 +407,14 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
             self.send_exception(mm_error)
 
     @BACKEND_ACTION_SNIPPET.time()
-    def send_snippet(self, text, title=None):
+    def send_snippet(self, text, title=None, ephemeral=None):
```
glennmatthews, line 410: If I'm reading the diff correctly, you're *only* adding `ephemeral` as an option to `MattermostDispatcher.send_snippet()` and `SlackDispatcher.send_snippet()`, but not to the other dispatchers? That seems incorrect to me - all dispatchers *must* have the same public API, even if they don't necessarily make use of certain optional parameters.

nautobot_chatops/dispatchers/slack.py
```diff
@@ -208,16 +213,36 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
+                            "type": "section",
+                            "text": {
+                                "type": "mrkdwn",
+                                "text": text
+                            }
+                        }
+                    ]
+                )
+            else:
+                self.slack_client.files_upload(channels=channels, content=text, title=title)
```
glennmatthews, line 245: The `len(text) < 40000` check seems like a potentially unwelcome surprise - if the chatbot is configured for `send_all_messages_private`, having messages over a certain length be unexpectedly sent as non-privately could be a security concern.

Maybe instead we could split the text into chunks and call `chat_postEphemeral` multiple times if needed?

nautobot_chatops/dispatchers/adaptive_cards.py
jvanderaa: This is a regression of the last time on it too. I had some challenges on my dev environment and missed getting this cleared up. I will remove the ephemeral part on this.

nautobot_chatops/dispatchers/slack.py
jvanderaa: The toughest part here is that the likely place to have the split is on a large table. If that is the case we may need to look at handling headers appropriately. 

I'll think about this one some more and how we may be able to handle it.

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -403,11 +407,14 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
         """Send a longer chunk of text as a file snippet."""
         channel = [self.context.get("channel_id")]
         logger.info("Sending snippet to %s: %s", channel, text)
-        self.mm_client.chat_post_message(channel_id=self.context.get("channel_id"), snippet=text)
+        if ephemeral:
+            self.mm_client.chat_post_ephemeral(channel_id=self.context.get("channel_id"), message=text)
```
smk4664, line 415: I don't agree with this line. If sending ephemeral, then it no longer sends as a snippet. This means that it is limited to the message size and could break some commands with larger tables.

jvanderaa: So a limiter would need to be setup? Looks like the max message size is 16,383 chars.

https://mattermost.com/blog/mattermost-5-0-intercept-and-modify-posts-advanced-permissions-longer-posts-and-more/#:~:text=Increased%20character%20limits%20on%20posts,better%20Markdown%20formatting%2C%20including%20tables.

jvanderaa: How would you want to handle this within the MM client @smk4664 ?

nautobot_chatops/dispatchers/slack.py
jvanderaa: Any recommendations/thoughts on how to have the split occur? I'd like to some how do something with splitting on a `\n` character. Perhaps something with a look ahead in a function? I may have something on that later...

nautobot_chatops/dispatchers/mattermost.py
smk4664: My concern is that this function is used specifically to send a snippet, but then you are removing that if it is ephemeral. I would add the snippet option to the chat_post_ephemeral.  I am not sure if you can, but it is worth testing.

jvanderaa: I think I have a solution in the latest.

jvanderaa:
Working to figure out the CI (some pylint disables coming where applicable).

docs/design.md
```diff
@@ -162,4 +162,6 @@ Some known limitations of currently supported platforms:
 - No table functionality in cards (blocks). The `ColumnSet` card layout feature allows for wrapping content across
   multiple columns, but does not provide for any alignment across columns, so it's not suitable for tables.
 
+## Settings
 
+The setting of `send_all_messages_private` within the configuration applied within the Nautobot config is used to send all messages as private messages. The messages will be sent in private when this is set to `True`. The default setting is `False`, which is the behavior for several message settings.
```
glennmatthews, line 167: Not sure I understand "which is the behavior for several message settings."?

Also it would be good to call out which platforms do and do not support this behavior - e.g. MS Teams and Webex simply ignore the `ephemeral` flag as that isn't AFAIK a capability the platform supports.

nautobot_chatops/dispatchers/slack.py
```diff
@@ -208,16 +213,30 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
         try:
-            self.slack_client.files_upload(channels=channels, content=text, title=title)
+            # Check for the length of the file if the setup is meant to be a private message
+            if ephemeral:
+                message_list = self.split_messages(text, 40000)
+                for msg in message_list:
+                    self.slack_client.chat_postEphemeral(
+                        channel=self.context.get("channel_id"),
+                        user=self.context.get("user_id"),
+                        text="test",
```
glennmatthews, line 235: Probably unneeded?

nautobot_chatops/dispatchers/slack.py
```diff
@@ -208,16 +213,30 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
+                        blocks=[{"type": "section", "text": {"type": "mrkdwn", "text": msg}}],
+                    )
```
glennmatthews, line 237: Why not just `self.send_blocks(self.markdown_block(msg), ephemeral=ephemeral)`?

nautobot_chatops/dispatchers/slack.py
jvanderaa: I think this is needed. This is how it is sent ephemerally.

glennmatthews: IIRC I was referring specifically to `text="test"` as being unneeded.

docs/design.md
```diff
@@ -109,24 +109,25 @@ every command as a top-level worker function. (`/nautobot get-device-info <devic
 rather than `/nautobot-get-device-info <device>`, `/nautobot-get-vlan-info <vlan>`, etc.) This is because:
 
 a. On platforms such as Slack, each separate slash-command must be enabled and configured separately on the server,
-   so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
+so an excessive number of distinct top-level commands will make the chatbot inconvenient to deploy.
 b. Platforms such as Microsoft Teams may limit the number of top-level commands that are displayed to the user in
-   a chat client, so large numbers of commands may be difficult to discover.
+a chat client, so large numbers of commands may be difficult to discover.
```
glennmatthews, line 114: This looks wrong to me - it is rendering as a paragraph instead of a list:

image

docs/design.md
```diff
@@ -157,9 +158,21 @@ Some known limitations of currently supported platforms:
...
+
+### Settings - Platform Support of Settings
+
+> This table represents the platform support of particular settings
 
+| Platform | send_all_messages_private |
+| Slack | ✅ |
+| MS Teams | ❌ |
+| WebEx | ❌ |
+| Mattermost | ✅ |
```
glennmatthews, line 178: This isn't rendering as a table - I think you need to add a `| -------- | ------------------------- |` after the initial line:

image

nautobot_chatops/dispatchers/adaptive_cards.py
```diff
@@ -121,7 +120,13 @@ def multi_input_dialog(self, command, sub_command, dialog_title, dialog_list):
...
         }
         blocks.append(buttons)
-        return self.send_blocks(blocks, callback_id=callback_id, modal=True, ephemeral=False, title=dialog_title)
+        return self.send_blocks(
+            blocks,
+            callback_id=callback_id,
+            modal=True,
+            ephemeral=False,
+            title=dialog_title,
+        )
```
glennmatthews, line 129: Unnecessary formatting-only change?

nautobot_chatops/dispatchers/base.py
```diff
@@ -327,3 +340,24 @@ def select_element(self, action_id, choices, default=(None, None), confirm=False
     def text_element(self, text):
         """Construct a simple plaintext element."""
         raise NotImplementedError
+
+    @staticmethod
+    def split_messages(text_string: str, max_message_size: int) -> list:
```
glennmatthews, line 345: Nit: perhaps `split_message` would be better?

nautobot_chatops/dispatchers/mattermost.py
```diff
@@ -403,11 +407,17 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
 
     @BACKEND_ACTION_SNIPPET.time()
-    def send_snippet(self, text, title=None):
+    def send_snippet(self, text, title=None, ephemeral=None):
         """Send a longer chunk of text as a file snippet."""
         channel = [self.context.get("channel_id")]
         logger.info("Sending snippet to %s: %s", channel, text)
-        self.mm_client.chat_post_message(channel_id=self.context.get("channel_id"), snippet=text)
+        if ephemeral:
+            for msg in self.split_messages(text, 16383):
```
glennmatthews, line 415: Can we make `16383` into a module-level or class-level constant so it's not just a magic number here?

nautobot_chatops/dispatchers/slack.py
```diff
@@ -208,16 +213,25 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
             channels = [self.context.get("user_id")]
         else:
             channels = [self.context.get("channel_id")]
         channels = ",".join(channels)
         logger.info("Sending snippet to %s: %s", channels, text)
         try:
-            self.slack_client.files_upload(channels=channels, content=text, title=title)
+            # Check for the length of the file if the setup is meant to be a private message
+            if ephemeral:
+                message_list = self.split_messages(text, 40000)
```
glennmatthews, line 230: Same request here about defining a constant.

nautobot_chatops/dispatchers/slack.py
```diff
@@ -208,16 +213,25 @@ def send_blocks(self, blocks, callback_id=None, modal=False, ephemeral=False, ti
...
+                for msg in message_list:
+                    self.send_blocks(self.markdown_block(msg), ephemeral=ephemeral)
```
glennmatthews, line 232: I don't think we want markdown formatting to apply to `send_snippet` in this case - is it possible to send it as plaintext, or at least wrap it in ```` ```\n...\n``` ```` (similar to `WebexDispatcher.send_snippet()`)?

tasks.py
```diff
@@ -41,12 +41,17 @@ def is_truthy(arg):
...
-            "nautobot_ver": "1.0.1",
+            "nautobot_ver": "1.1.6",
             "project_name": "nautobot-chatops",
             "python_ver": "3.6",
             "local": False,
             "compose_dir": os.path.join(os.path.dirname(__file__), "development"),
-            "compose_files": ["docker-compose.requirements.yml", "docker-compose.base.yml", "docker-compose.dev.yml"],
+            "compose_files": [
+                "docker-compose.requirements.yml",
+                # "docker-compose.celery.yml",
```
glennmatthews, line 51: If 1.1 is now the default nautobot version, don't we need to include celery by default?

jvanderaa: It will be. I'm going to do the update to nautobot dev environment in a separate PR. This is a better method to have the Celery components commented out but in the code. It is just easier then to get it up and running.

jvanderaa: This is just an easier method to handle converting over to using 1.1 or later. When Nautobot 1.3 comes out we will make the updates to go to Python3.7 and update to the newer methods.

glennmatthews (approved):

PR #128, opened 2022-03-08
QOL tweaks / bugfix

matt852:
I've added two minor changes into this PR:

The header for the page for adding a new access grant, as well as a new command token, is supposed to change between "Adding" and "Editing".  However this is always showing as "Editing".  The reason is the `obj.pk` is generated when the page is loaded, so it always appear to exist.

The second item is a quality of life change, but it makes autofocuses the primary text field when opening the "Access Grant" or "Command Token" form.  When doing this, I moved the help text strings to a separate, static file to prevent duplicating their values across models.py and forms.py.

### DIFF
--- a/nautobot_chatops/templates/nautobot/access_grant_edit.html
+++ b/nautobot_chatops/templates/nautobot/access_grant_edit.html
@@ -2,7 +2,7 @@
 {% load static %}
 {% load form_helpers %}
 
-<h1>{% block title %}{% if obj.command %}Editing{% else %}Add a new{% endif %} Nautobot Access Grant{% endblock %}</h1>
+<h1>{% block title %}{% if obj.present_in_database %}Editing{% else %}Add a new{% endif %} Nautobot Access Grant{% endblock %}</h1>
 
 {% block form %}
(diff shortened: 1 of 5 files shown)


nautobot_chatops/templates/nautobot/access_grant_edit.html
```diff
@@ -2,7 +2,7 @@
 {% load static %}
 {% load form_helpers %}
 
-<h1>{% block title %}{% if obj.pk %}Editing{% else %}Add a new{% endif %} Nautobot Access Grant{% endblock %}</h1>
+<h1>{% block title %}{% if obj.command %}Editing{% else %}Add a new{% endif %} Nautobot Access Grant{% endblock %}</h1>
```
glennmatthews, line 5: IIRC this and the other below can be changed to `if obj.present_in_database`.

matt852: Yup that worked.  I changed both.  Thank you.

glennmatthews (approved):

Issue #127, opened 2022-03-08
Hidden message option causes error when receiving a response


david-kn:
### Environment
* Nautobot version:  1.2.7
* nautobot-plugin-chatops version:  1.6.0

### Steps to Reproduce
1. Enable settings `"send_all_messages_private": True` in plugin configuration
2. Type any `/nautobot` command in Slack
3. Response from Slack is an error message
![image](https://user-images.githubusercontent.com/14314517/157216761-5ea6ff23-10df-401a-99ee-94fb14586399.png)

### Expected Behavior
Data should be returned 'normally' (via ephemereal (hidden) message).

### Observed Behavior
This issue happened only when private message option is enabled.

ubajze:
The problem in `1.7.0` is described in #132 .

jvanderaa:
Fixed in #133

PR #129, opened 2022-03-09
Fixes Private Message Error

jvanderaa:
The original implementation missed sending one section as blocks. This PR fixes that reported in #127.

During attempting to resolve this, I also found that we had the logger set to the rq.worker only. This has been updated to solve the issue identified in 127.

### DIFF
--- a/nautobot_chatops/dispatchers/slack.py
+++ b/nautobot_chatops/dispatchers/slack.py
@@ -232,7 +232,7 @@ class SlackDispatcher(Dispatcher):
                 message_list = self.split_message(text, SLACK_MAX_MESSAGE_LENGTH)
                 for msg in message_list:
                     # Send the blocks as a list, this needs to be the case for Slack to send appropriately.
-                    self.send_blocks([self.markdown_block(msg)], ephemeral=ephemeral)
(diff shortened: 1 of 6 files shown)


nautobot_chatops/dispatchers/slack.py
```diff
@@ -231,7 +231,8 @@ def send_snippet(self, text, title=None, ephemeral=None):
             if ephemeral:
                 message_list = self.split_message(text, SLACK_MAX_MESSAGE_LENGTH)
                 for msg in message_list:
-                    self.send_blocks(self.markdown_block(f"```\n{msg}\n```"), ephemeral=ephemeral)
+                    # Send the blocks as a list, this needs to be the case for Slack to send appropriately.
+                    self.send_blocks([self.markdown_block(msg)], ephemeral=ephemeral)
```
glennmatthews, line 235: Don't we still need to have the f-string to apply code fencing to `msg`?
```suggestion
                    self.send_blocks([self.markdown_block(f"```\n{msg}\n```")], ephemeral=ephemeral)
```

glennmatthews (approved):

PR #133, opened 2022-04-07
Fix slack msg limits

jvanderaa:
- @ubajze found that the message length limit on ephemeral messages is 3000 chars
- Sets up to have a default setting to allow for t-shooting if necessary and lower the size

### DIFF
--- a/nautobot_chatops/__init__.py
+++ b/nautobot_chatops/__init__.py
@@ -49,7 +49,7 @@ class NautobotChatOpsConfig(PluginConfig):
         # As requested on https://github.com/nautobot/nautobot-plugin-chatops/issues/114 this setting is used for
         # sending all messages as an ephemeral message, meaning only the person interacting with the bot will see the
         # responses.
-        "send_all_messages_private": False,"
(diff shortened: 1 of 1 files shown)


nautobot_chatops/dispatchers/slack.py
```diff
@@ -23,7 +23,7 @@
 BACKEND_ACTION_BLOCKS = backend_action_sum.labels("slack", "send_blocks")
 BACKEND_ACTION_SNIPPET = backend_action_sum.labels("slack", "send_snippet")
 
-SLACK_MAX_MESSAGE_LENGTH = 40000
```
glennmatthews, line 26: Was SLACK_MAX_MESSAGE_LENGTH really not used anywhere else? That surprises me.

glennmatthews (approved):

nautobot_chatops/dispatchers/slack.py
jvanderaa: Yeah, I basically changed the name of the constant variable to reflect that more proper.

PR #134, opened 2022-04-19
Nautobot 1.3 Support

nniehoff:
This PR:

Removes Python 3.6 support
Removes Nautobot 1.0 support
Adds Nautobot 1.3 support
Removes Celery as an explicit dependency and adds Nautobot >= 1.1.0 as a real dependency


.github/workflows/ci.yml
```diff
@@ -113,9 +114,29 @@ jobs:
...
+        if: "matrix.nautobot-version != '1.3.0'"
+        run: "poetry remove black --dev"
+      - name: "Handle Markupsafe bug"
+        # Markupsafe 2.0.1 conflicts with the new jinja, but 2.1.x conflicts with Nautobot 1.1-2
+        # Markdown has issues with importlib in newer versions
+        if: "matrix.nautobot-version != '1.3.0'"
+        run: "poetry add markupsafe==2.0.1"
+      - name: "Set Nautobot install to the testing version"
+        run: "poetry add nautobot==${{ matrix.nautobot-version }}"
+      - name: "Remove Poetry Lock for Older Nautobot Tests"
```
glennmatthews, line 137: Perhaps?

```suggestion
      - name: "Regenerate Poetry Lock for Older Nautobot Tests"
```

Or?

```suggestion
      - name: "Refresh Poetry Environment for Older Nautobot Tests"
```

.github/workflows/ci.yml
```diff
@@ -113,9 +114,29 @@ jobs:
...
+        run: "poetry update --lock"
```
glennmatthews, line 138: I don't really understand why this step is needed, but I believe you that it is. :-)

glennmatthews (approved):

.github/workflows/ci.yml
nniehoff: I tend to agree with you but looking back at the GHA logs this step is actually doing stuff with the lock file so I'm hesitant to remove it...

pyproject.toml
```diff
@@ -17,21 +17,17 @@ include = [
 "nautobot" = "nautobot_chatops.workers.nautobot:nautobot"
 
 [tool.poetry.dependencies]
-python = "^3.6"
-nautobot = "*"
+python = "^3.7"
+nautobot = "^1.1.0"
```
smk4664, line 21: I thought this was pattern was against plugin policy as it can automatically upgrade a Nautobot instance.  That said, probably a safe value.

.github/workflows/ci.yml
smk4664: I am curious, instead of adding nautobot in the pyproject.toml, could we instead do it here?

pyproject.toml
smk4664: Testing, this doesn't seem to break my nautobot instance. I am good with this change.

smk4664 (changes requested):
Can you update the min_version in the __init__.py?

nautobot_chatops/__init__.py

smk4664 (approved):
LGTM

PR #159, opened 2022-10-06
Add Compatibility Matrix

smk4664:
This PR adds a Compatibility Matrix to show which versions of the Chatops plugin are compatible with the different Nautobot versions.


whitej6 (commented):
For 1.9.X are we staying with 1.5.99 or based on recent discussions internally go to 1.99?

smk4664:
This aligns with what Golden Config has, plus a little birdie told me that 1.5 is the last Nautobot 1.x release.

PR #165, opened 2022-11-15
Update MS Teams ZIP file

matt852:
Updates the MS Teams ZIP file with a couple new settings from Azure/Teams.


smk4664:
I compared the manifest between the existing zip and the new zip. I noticed that `configurableProperties` was removed. Was this an intentional change?

matt852:
> I compared the manifest between the existing zip and the new zip. I noticed that `configurableProperties` was removed. Was this an intentional change?

Yes.  This ZIP file was automatically generated from the Developer Portal, which seems to have removed these settings.  I believe they are optional, so I left them excluded.

PR #171, opened 2022-12-05
Enable Upstream Testing

bryanculver:
# Closes: #DNE

## What's Changed
- Updated Dockerfile to be compatible with upstream testing
- Updated invoke, Dockerfile defaults which are incompatible with `pyproject.toml`
- Added workflow fragment to trigger testing

## TODO

- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/chatops/en/latest/dev/dev_contributing/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Outline Remaining Work, Constraints from Design


.github/workflows/upstream_testing.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: "Nautobot Upstream Testing"
+
+on: # yamllint disable-line rule:truthy rule:comments
+  schedule:
+    - cron: "0 4 */2 * *"  # every other day at midnight
+  push:  # yamllint disable rule:empty-values
+# TODO: REMOVE PUSH LINE ABOVE
```
bryanculver, line 8:
```suggestion
```
Remove before merging.

bryanculver:
`next` is expected to fail because it goes to a new major version of Nautobot so can be safely ignored for now.

tasks.py
```diff
@@ -39,7 +39,7 @@ def is_truthy(arg):
 namespace.configure(
     {
         "nautobot_chatops": {
-            "nautobot_ver": "1.1.6",
+            "nautobot_ver": "1.2.0",
```
bryanculver, line 42: This is a mismatch from the `pyproject.toml`: https://github.com/nautobot/nautobot-plugin-chatops/blob/1dcb3cf024efc70fd2f604b3be5d11fdc9938cb3/pyproject.toml#L25

bryanculver:
`pylint` failure is exposing:

The `PluginConfig` is a mismatch between the `pyproject.toml` and the invoke defaults:

* https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops/__init__.py#L61
* https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/pyproject.toml#L25
* https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/tasks.py#L42

.github/workflows/upstream_testing.yml
jvanderaa: Makes sense.

tasks.py
jvanderaa: Agreed

bryanculver:
Tried to use `poetry self update` to deal with `nautobot-dev` containers potentially having an older version of Poetry where CI/dev might have a newer one and mint a lock file that is too new (see: https://github.com/networktocode-llc/cookiecutter-ntc/pull/276#discussion_r974383161, https://github.com/nautobot/nautobot-plugin-chatops/actions/runs/4088225247/jobs/7049757009#step:7:194).

smk4664 (approved):
Thanks Bryan for fixing the CI and improving the process!  Hopefully in the future we can just use poetry instead of the pip install with constraints.

bryanculver:
> 

Until something like this comes in: https://github.com/python-poetry/poetry/issues/7051 I don't think we can go all-in for Poetry for this use case. We need to test a package within an environment of already installed packages. We don't need to be hand-crafting multiple lock files or `pyproject.toml`s.

Issue #174, opened 2022-12-27
Legacy list templates need to be changed to extend `generic/object_list.html`


HanlinMiao:
https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops/templates/nautobot/access_grant_list.html
https://github.com/nautobot/nautobot-plugin-chatops/blob/develop/nautobot_chatops/templates/nautobot/command_token_list.html

These two list templates need to be changed to extend `generic/object_list.html`

PR #175, opened 2023-01-30
Changed extends base.html to extends generic/object_list.html for list templates

briddo:
# Closes: #174 

## What's Changed
Changed `extends 'base.html'` to `extends 'generic/object_list.html'` in list templates and removed unneeded `load`s from these templates.


glennmatthews:
How much of `block content` can now be removed after changing the base template?

smk4664 (approved):
Thanks for your help @briddo

Issue #187, opened 2023-03-25
Dev Updates


jvanderaa:
### Environment
* Nautobot version:  
* nautobot-plugin-chatops version:  

### Proposed Functionality

Add watchmedo to celery containers

### Use Case

Not having to reload the celery worker when a change happens.

PR #188, opened 2023-03-25
Dev Updates

jvanderaa:
# Closes: #187

## What's Changed

- Adds dependabot.yml config based on @bryanculver work
- Adds watchmedo to the dev container for celery
- Disables the alert of WEBEX token missing when the enable_webex is false

## TODO

- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/chatops/en/latest/dev/dev_contributing/#creating-changelog-fragments))
- [x] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests
- [x] Outline Remaining Work, Constraints from Design


development/docker-compose.base.yml
```diff
@@ -21,6 +21,8 @@ services:
     depends_on:
       - "postgres"
       - "redis"
+    healthcheck:
```
itdependsnetworks, line 24: I was just thinking about this... I think we should do this, but it should be in `docker-compose.dev.yml` instead.

jvanderaa: I can get this moved, but brings up another question, what is the point of the `docker-compose.dev.yml` compared to `docker-compose.yml` within the `development` directory? Is it a bit too many files then, since everything in the directory should be a development file. 

It's just a question that I really don't have an answer for.

dependabot.yml
cmsirbu, line 1: Shouldn't this be in the `.github` folder?

jvanderaa:
Too complex of updates. Closing.

PR #209, opened 2023-05-26
Add Ansible Integration

snaselj:
# Closes: #202 

## What's Changed

- Merged nautobot-plugin-chatops-ansible to nautobot-plugin-chatops.


snaselj:
This PR is ready for review now.

development/nautobot_config.py
```diff
@@ -133,6 +133,12 @@ def _get_bool_env(name: str):
 # Each key in the dictionary is the name of an installed plugin and its value is a dictionary of settings.
 PLUGINS_CONFIG = {
     "nautobot_chatops": {
+        "ansible": {
+            "tower_uri": os.getenv("NAUTOBOT_TOWER_URI"),
+            "tower_username": os.getenv("NAUTOBOT_TOWER_USERNAME"),
+            "tower_password": os.getenv("NAUTOBOT_TOWER_PASSWORD"),
+            "tower_verify_ssl": _get_bool_env("NAUTOBOT_TOWER_VERIFY_SSL", True),
+        },
```
smk4664, line 141: Could we remove the nesting here?
```suggestion
        "ansible_tower_uri": os.getenv("NAUTOBOT_TOWER_URI"),
        "ansible_tower_username": os.getenv("NAUTOBOT_TOWER_USERNAME"),
        "ansible_tower_password": os.getenv("NAUTOBOT_TOWER_PASSWORD"),
        "ansible_tower_verify_ssl": _get_bool_env("NAUTOBOT_TOWER_VERIFY_SSL", True),
```

pyproject.toml
```diff
@@ -18,17 +18,21 @@ include = [
...
 "clear" = "nautobot_chatops.workers.clear:clear"
 "nautobot" = "nautobot_chatops.workers.nautobot:nautobot"
+"ansible" = "nautobot_chatops.integrations.ansible.worker:ansible"
 
 [tool.poetry.dependencies]
 # TODO Move to ~ requirements instead of ^
 python = "^3.7"
 Markdown = "!=3.3.5"
 PyJWT = "^2.1.0"
+PyYAML = "*"
```
smk4664, line 28: I don't think we should be allowing all versions of PyYAML, but I would be curious to hear what others would say?

Also, shouldn't these three go in the extras only? If they are in the [tool.poetry.dependencies] it will still install them.

smk4664 (changes requested):
I am curious what happens when you don't install the extras? This one doesn't seem like it would have an issue as I am pretty sure Pydantic isn't used for the Ansible Integration, and PyYAML/requests are both going to come in from Nautobot. But what about other integrations?

pyproject.toml
```diff
@@ -18,17 +18,21 @@ include = [
...
 aiodns = "^1.0"
 aiohttp = "^3.7.3"
 asgiref = "^3.4.1"
 nautobot = "^1.4.0"
 nautobot-capacity-metrics = "*"
+pydantic = "^1.8.2"
+requests = "^2.25.1"
```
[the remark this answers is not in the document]
snaselj: IMHO it's a standard practice to specify all direct dependencies. We should make version wide enough to cover compatibility with the dependency, and keep it installable with Nautobot. Will check it again.

pyproject.toml
snaselj: Will fix extra dependencies to be compatible with https://python-poetry.org/docs/pyproject/#extras

nautobot_chatops/integrations/ansible/tower.py
```diff
@@ -0,0 +1,222 @@
...
+    return valid_uri.geturl().rstrip("/")
+
+
+class Tower:  # pylint: disable=too-many-function-args
+    """Representation and methods for interacting with Ansible Tower/AWX."""
+
+    def __init__(
+        self,
+        origin,
+        tower_uri=os.getenv("NAUTOBOT_TOWER_URI"),
```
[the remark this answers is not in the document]
snaselj: Fixed

nautobot_chatops/integrations/ansible/worker.py
```diff
@@ -0,0 +1,251 @@
...
+import os
+from collections import namedtuple
+
+import yaml
+from django_rq import job
+from nautobot_chatops.workers import handle_subcommands, subcommand_of
+
+from .tower import Tower
+
+TOWER_URI = os.getenv("NAUTOBOT_TOWER_URI")
```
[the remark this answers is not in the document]
snaselj: Fixed

pyproject.toml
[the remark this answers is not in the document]
snaselj: Fixed

`requests` is missing in direct Nautobot deps, its only in dev deps there. Versions match with Nautobot >=1.4.0.

Removed pydantic as unused.

development/nautobot_config.py
snaselj: Fixed.

pyproject.toml
snaselj: Fixed.

development/nautobot_config.py
bryanculver: While we're at it, can we make the OS environment variable more descriptive? `NAUTOBOT_CHATOPS_ANSIBLE_TOWER_URI`? I know it's long but can we cut down on overlap?

snaselj: Unifying environment variable names is a good point. I already changed some of it (e.g. `ENABLE_MATTERMOST` => `NAUTOBOT_CHATOPS_ENABLE_MATTERMOS`), and it brought some confusion to @pke11y . Will keep track of this issue for the later time.

smk4664 (approved):
Thanks for your contributions! This looks great!

PR #265, opened 2023-09-25
Align Integrations Documentation Structure

snaselj:
# Closes: NaN

## What's Changed

- Aligned integrations documentation structure with SSoT.

PR #288, opened 2023-10-27
Add NSO integration

romanukes:
TODO:

- [x] Move tests
- [x] Check for compatibility with Nautobot 2.0 (Remove device_role, slugs, etc.)
- [x] Add CommandFilterListView and Navigation. Disable view if NSO is disabled.


cmsirbu:
Quick note to say that integration docs structure has been slightly changed by #265 (to align with the way it's done by SSoT), which will require a bit of an update in this PR and a rebase.

smk4664 (commented):
I am curious what CommandFilter is used for. Is it not possible to do something similar without adding a new model?

nautobot_chatops/migrations/0008_auto_20231228_2231.py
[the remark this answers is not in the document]
romanukes: done

nautobot_chatops/integrations/nso/models.py
smk4664, line 1: As discussed on call, we are going to try to replace this model with a config instead, so we do not add a required model in the database for an optional integration.

smk4664 (changes requested):
As discussed, we will be removing the model and switching to another method for filtering commands.

development/mattermost/nautobot_bootstrap.py
```diff
@@ -48,7 +49,7 @@
 with contextlib.suppress(ObjectDoesNotExist):
     admin = User.objects.get(name="admin")
     ChatOpsAccountLink.objects.update_or_create(
-        user_id="jactwicuqb8bu8pau8mgjydzeo",
```
smk4664, line 51: Why did this change?

romanukes: To match the Mattermost SQL dump. Mattermost integration didn't work without this change

smk4664 (approved):
Thank you for all of your work for this!

PR #293, opened 2023-12-20
Add debugs to ms teams. Fix Aristacv errors.

zackt25:
MS Teams files doesn't really have any debug messages for helping troubleshoot so I added some. 

I also fixed a integration issue with Aristacv and cloudvision.


whitej6:
@zackt25 I'd recommend splitting into two PRs. Debug as one PR and a separate PR for renaming. For the renaming there's a lot more work that would be needed and to do it properly consistent branding the rename would need to a major release as it would change how to configure ChatOps for cloud vision.

smk4664:
> @zackt25 I'd recommend splitting into two PRs. Debug as one PR and a separate PR for renaming. For the renaming there's a lot more work that would be needed and to do it properly consistent branding the rename would need to a major release as it would change how to configure ChatOps for cloud vision.

The renaming piece is complicated. The original name of the command was "Cloudvision". When combined in the integration, we tried to rename it to Arista CloudVision to match the SSOT. But due to the way that the commands are picked up by the workers, the folder name must match the command. With what is in place, the Cloudvision plugin tries to load, even if you have cloudvision disabled. 

https://github.com/nautobot/nautobot-app-chatops/blob/5a4f3548435a860f02e4be5f0f3d6e5b9b609504/nautobot_chatops/integrations/utils.py#L12-L16
Is the problematic line that uses the folder name.

PR #299, opened 2024-02-05
Updates codeowners.

jvanderaa:
Removes @jvanderaa  from codeowners.


whitej6 (approved):
We need to move everything to groups but can handle in a separate PR

smk4664 (approved):

PR #311, opened 2024-04-14
Update Account Link docs to refer to fallback_chatops_user option

itdependsnetworks:
Update Account Link docs to refer to fallback_chatops_user option


smk4664 (approved):
Thanks for the doc additions!

PR #315, opened 2024-06-13
Pin Meraki to max 1.45.0

smk4664:
Pin Meraki to 1.45.0 due to Meraki removing support for Python <3.10 in 1.46.0


pyproject.toml
```diff
@@ -54,7 +54,7 @@ ipaddr = { version = "^2.2.0", optional = true }
 ipfabric = { version = "~6.0.9", optional = true }
 ipfabric-diagrams = { version = "~6.0.2", optional = true }
 isodate =  { version = "^0.6.1", optional = true }
-meraki = { version = "^1.7.2", optional = true }
+meraki = { version = "<=1.45.0", optional = true }
```
glennmatthews, line 57: Want a minimum version here still I assume?

smk4664: Ah, good point. I will add in.

glennmatthews (approved):

PR #323, opened 2024-08-23
Replace linters with Ruff

smk4664:
Upstream testing is failing in ltm-1.6 due to a conflict between pycodestyle and a dependency in the Nautobot container. This change removes Bandit, Black, Flake8 and PyDocStyle in favor of Ruff. This removes the conflict with pycodestyle and also speeds up our testing.


smk4664:
This didn't work, going to just move to ruff.

bryanculver (approved):

PR #338, opened 2024-10-17
Remove config file from migrations check

smk4664:
We should not be running with the test config file as we want the Chatops plugin loaded.

### DIFF
--- a/changes/338.housekeeping
+++ b/changes/338.housekeeping
@@ -1,0 +1,1 @@
+Remove --config option from make-migrations check.
--- a/changes/338.housekeeping
+++ b/changes/338.housekeeping
@@ -1,1 +1,1 @@
-Remove --config option from make-migrations check.
+Fixed `invoke check-migrations`.
--- a/tasks.py
+++ b/tasks.py
@@ -591,7 +591,7 @@ def yamllint(context):
 @task
 def check_migrations(context):
     """Check for missing migrations."""
-    command = "nautobot-server --config=nautobot/core/tests/nautobot_config.py makemigrations --dry-run --check"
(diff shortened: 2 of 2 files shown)


tasks.py
```diff
@@ -591,7 +591,7 @@ def yamllint(context):
 @task
 def check_migrations(context):
     """Check for missing migrations."""
-    command = "nautobot-server --config=nautobot/core/tests/nautobot_config.py makemigrations --dry-run --check"
+    command = "nautobot-server makemigrations --dry-run --check"
```
glennmatthews, line 594: Was this specific to this app or is it something that needs to be fixed more generally?

glennmatthews (approved):

tasks.py
smk4664: This was pre-drift manager. This is fixed on drift-manager, but likely copy/pasted from another repo. I found this in Welcome Wizard and I will be checking other repos.

gsnider2195: It appears to be a problem in the ltm branch of multiple apps, perhaps all of them

gsnider2195: We're not drift managing ltm so this may have to be manually fixed on every app as we start fixing upstream testing

changes/338.housekeeping
```diff
@@ -0,0 +1 @@
+Remove --config option from make-migrations check.
```
gsnider2195, line 1:
```suggestion
Fixed `invoke check-migrations`.
```

tasks.py
glennmatthews: Ack - I missed this was an LTM fix.

PR #339, opened 2024-10-21
Update .cookiecutter.json

smk4664:
Due to the way that Drift Manager uses the .cookiecutter.json file, we need to change the model_class_name to a valid model in BGP models to help us track drift in files that would be removed if the model_class_name=None.

### DIFF
--- a/changes/339.housekeeping
+++ b/changes/339.housekeeping
@@ -1,0 +1,1 @@
+Change model_class_name in .cookiecutter.json to a valid model.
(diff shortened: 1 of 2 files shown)


romanukes (approved):

changes/339.housekeeping
```diff
@@ -0,0 +1 @@
+Change model_class_name in .cookiecutter.json to a valid model.
```
smk4664, line 1:
```suggestion
Changed model_class_name in .cookiecutter.json to a valid model to help with drift management.
```

PR #341, opened 2024-11-13
Fix constance being accessed before database is ready and other fixes

### LINKED ISSUE #340: Constance is being accessed before database is ready
This is preventing core from upgrading django-constance to v4 because chatops prevents the app from starting when it tries to access the database before it's ready. See [here](https://github.com/nautobot/nautobot-app-chatops/blob/757ed3e921e45cc742f6cb938373db0893f25ec4/nautobot_chatops/urls.py#L28)

gsnider2195:
# Closes #340

## What's Changed

- Remove conditional logic for urlpatterns based on constance settings
- Update existing API views that were affected by the above change to return a 404 if the integration is not enabled
- Remove conditional logic for adding grafana navigation menu items
  - We can't do this at init because it needs to access the database
  - We can't modify the nav menu during runtime because core doesn't support removing nav menu entries dynamically
- Add a "grafana disabled" view in case a user clicks on a grafana nav menu item when the grafana integration is disabled
- Fix multiple bugs in the grafana views
  - Incorrect view names
  - Incorrectly using a reference to the model instead of the model itself
  - Delete all grafana API files since there we no API views provided by grafana integration
  - Fixed Nautobot v2.3 incompatibility caused by saved views not being able to determine the models' table classes
- Fix dev environment nautobot_config.py to fall back to constance if environment variable is not used
- Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in
- Add exception handling for cases when diffsync is not installed, since it's marked as optional

## Screenshots

image


glennmatthews:
I'm curious about:
> Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in

These are the endpoints for receiving messages *from* the chat platform, aren't they? I wouldn't expect them to require Nautobot user authentication.

nautobot_chatops/integrations/grafana/diffsync/sync.py
```diff
@@ -22,6 +28,9 @@ def run_dashboard_sync(overwrite: bool = False) -> Union[str, None]:
     Args:
         overwrite (bool): Overwrite Nautobot data and delete records that are no longer in Grafana.
     """
+    if not DIFFSYNC_INSTALLED:
+        raise ImportError("DiffSync is not installed. Please install DiffSync to use the grafana integration.")
```
glennmatthews, line 32: ImportError for these calls feels weird to me - shouldn't it be something like RuntimeError?

nautobot_chatops/integrations/grafana/views.py
```diff
@@ -33,21 +35,41 @@
...
+
+    def dispatch(self, request, *args, **kwargs):
+        """Dispatch method for Grafana views."""
+        if not request.user.is_authenticated:
+            return self.handle_no_permission()
+        if not get_app_settings_or_config("nautobot_chatops", "enable_grafana"):
+            return TemplateResponse(
+                request=self.request,
+                template="nautobot_chatops_grafana/grafana_disabled.html",
+                context={},
```
gsnider2195, line 61: This should probably be a redirect to another view but this works.

smk4664:
> I'm curious about:
> 
> > Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in
> 
> These are the endpoints for receiving messages _from_ the chat platform, aren't they? I wouldn't expect them to require Nautobot user authentication.

Yeah, these need to be reverted. The API for these views is for the chat platform to communicate with Nautobot. Each of these views have their own way to authenticate the system. For instance, Slack signs their requests.

nautobot_chatops/integrations/grafana/diffsync/sync.py
gsnider2195: I was going to use RuntimeError but then I decided to replicate the exception you would get if you tried to import diffsync in this function. It seemed more appropriate to use ImportError here but I'm happy to change it to RuntimeError if you think that's better here.

gsnider2195:
> > I'm curious about:
> > > Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in
> > 
> > 
> > These are the endpoints for receiving messages _from_ the chat platform, aren't they? I wouldn't expect them to require Nautobot user authentication.
> 
> Yeah, these need to be reverted. The API for these views is for the chat platform to communicate with Nautobot. Each of these views have their own way to authenticate the system. For instance, Slack signs their requests.

That's simple enough to do. I didn't have the context for why these were left wide open. These should probably also be changed to rest_framework views at some point instead of django views?

nautobot_chatops/integrations/grafana/diffsync/sync.py
glennmatthews: Would be my preference, but I don't feel strongly about it.

Might be good either way to add a `Raises:` section to the docstring documenting this behavior?

nautobot_chatops/urls.py
```diff
@@ -25,19 +23,6 @@
...
 )
 
-if get_app_settings_or_config("nautobot_chatops", "enable_grafana"):
-    try:
-        from nautobot_chatops.integrations.grafana.urls import urlpatterns as grafana_urlpatterns
-    # pylint: disable-next=broad-except
-    except Exception:
-        grafana_urlpatterns = []
-        logger = logging.getLogger(__name__)
-        logger.warning("Grafana ChatOps integration is not available.", exc_info=True)
```
glennmatthews, line 35: Wondering - were there cases where importing the grafana URLs might fail here (missing optional dependency, etc.) that we need to handle still?

glennmatthews:
> > > I'm curious about:
> > > > Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in
> > > 
> > > 
> > > These are the endpoints for receiving messages _from_ the chat platform, aren't they? I wouldn't expect them to require Nautobot user authentication.
> > 
> > 
> > Yeah, these need to be reverted. The API for these views is for the chat platform to communicate with Nautobot. Each of these views have their own way to authenticate the system. For instance, Slack signs their requests.
> 
> That's simple enough to do. I didn't have the context for why these were left wide open. These should probably also be changed to rest_framework views at some point instead of django views?

They don't use DRF serializers AFAIK so I'm not sure why they would be changed over?

gsnider2195:
> > > > I'm curious about:
> > > > > Fix mattermost, msteams, slack and webex integration API views that were not requiring user to be logged in
> > > > 
> > > > 
> > > > These are the endpoints for receiving messages _from_ the chat platform, aren't they? I wouldn't expect them to require Nautobot user authentication.
> > > 
> > > 
> > > Yeah, these need to be reverted. The API for these views is for the chat platform to communicate with Nautobot. Each of these views have their own way to authenticate the system. For instance, Slack signs their requests.
> > 
> > 
> > That's simple enough to do. I didn't have the context for why these were left wide open. These should probably also be changed to rest_framework views at some point instead of django views?
> 
> They don't use DRF serializers AFAIK so I'm not sure why they would be changed over?

Just so they don't have to manually disable csrf checks. This is the first place I've seen `@method_decorator(csrf_exempt, name="dispatch")` on an API view.

changes/341.housekeeping
```diff
@@ -0,0 +1 @@
+Fixed dev environment nautobot_config.py to fall back to constance if environment variable is not used.
```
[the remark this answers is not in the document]
gsnider2195: I forgot to add them. Thanks for catching that.

nautobot_chatops/urls.py
gsnider2195: Probably. Our dockerfile installs with `--extras all`. I'll test without that.

gsnider2195: Looks like I need to add more import guardrails 😞 

If I just put this exception handling back in we get a broken nav menu.

image

gsnider2195: This is fixed now. Discussed the implementation in DSU and decided not to spend too much time on it since ideally nobody will ever encounter this.

nautobot_chatops/models.py
```diff
@@ -23,9 +23,7 @@
     COMMAND_TOKEN_COMMENT_HELP_TEXT,
     COMMAND_TOKEN_TOKEN_HELP_TEXT,
 )
-from .integrations.grafana.models import Dashboard as GrafanaDashboard
-from .integrations.grafana.models import Panel as GrafanaPanel
-from .integrations.grafana.models import PanelVariable as GrafanaPanelVariable
+from .integrations.grafana.models import GrafanaDashboard, GrafanaPanel, GrafanaPanelVariable
```
smk4664, line 26: Oh good, I am glad that we can get rid of the rename. I am not sure why this wasn't done when originally implemented.

gsnider2195: I had to leave it in for the diffsync code because the diffsync models are called `GrafanaDashboard` and `GrafanaPanel` and I didn't want to go down another rabbit hole trying to fix those...

smk4664 (approved):
Thank you for the fixes!

PR #421, opened 2026-01-06
Slack token rotation

gioccher:
# Closes: #<ISSUE NUMBER GOES HERE>

## What's Changed
Added support for slack automatic token rotation.  
Token rotation is described here https://docs.slack.dev/authentication/using-token-rotation/ but the summary is that it replaces the static bot token with a pair of access and refresh tokens. Access expires every 12 hours, refresh is one-time use.  
Some companies security policies require token rotation, and this PR makes nautobot chatops work for them.  

## To Do

- [x] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [ ] Attached Screenshots, Payload Example
- [ ] Unit, Integration Tests
- [ ] Outline Remaining Work, Constraints from Design


smk4664 (approved):
Thanks for your work on this feature! This will be in the next release of Nautobot ChatOps.

PR #445, opened 2026-07-29
Updated aristacv extra for Nautobot 3.x / protobuf 5 compatibility

KamilLelonek:
> **Draft:** the dependency change alone is not sufficient - it forces a CloudVision tag API migration (below) whose runtime behavior needs validation against a live CloudVision instance before this should merge.

### What

Restore installability and Nautobot 3.x compatibility for the `aristacv` (Arista CloudVision) extra:

- `cloudvision`: `^1.1` -> `^1.25` (first CloudVision release built on protobuf 5)
- `protobuf`: `^3.17` (`<4`) -> `>=5.0,<6.0`
- regenerate `poetry.lock`
- migrate `cvpgrpcutils.get_device_tags` from the `arista.tag.v1` API to `arista.tag.v2`

### Why

The `aristacv` extra carried Nautobot 2.x-era pins. `protobuf<4` conflicts directly with Nautobot 3.x, which requires protobuf 5 (via `opentelemetry-exporter-otlp`), so `nautobot-chatops[aristacv]` can't be installed on Nautobot 3.x at all. CloudVision moved to protobuf 5 at `1.25.0` - but that same release **dropped `arista.tag.v1`** (now `arista.tag.v2`), which the CloudVision integration imported. So the dependency bump necessarily requires porting the tag lookup to v2.

v2 folds device tags into a general element-typed model: `DeviceTagAssignment*` becomes `TagAssignment*` with an explicit `element_type` (set to `ELEMENT_TYPE_DEVICE`). The `device_id`, `label`, and `value` fields are unchanged, so response parsing is untouched.

### Validation still needed (why this is a draft)

Verified so far: dependency resolution against Nautobot 3.2.1, `poetry.lock` consistency, and lint/pylint/ruff all pass.

**Not yet verified** - needs a live CloudVision instance:
- [ ] `get_device_tags` returns the same results against v2 as it did against v1
- [ ] Whether `workspace_id` must be set on the v2 key to read mainline (active) tag assignments - it is currently left unset (v1 had no workspace concept); if v2 requires it to see mainline state, this query may return nothing
- [ ] End-to-end smoke test of the CloudVision chatbot commands


glennmatthews:
```
************* Module nautobot_chatops.integrations.cloudvision.cvpgrpcutils
nautobot_chatops/integrations/cloudvision/cvpgrpcutils.py:5:0: E0401: Unable to import 'arista.tag.v1' (import-error)
nautobot_chatops/integrations/cloudvision/cvpgrpcutils.py:5:0: E0611: No name 'v1' in module 'arista.tag' (no-name-in-module)
```

glennmatthews:
Looks like `arista.tag.v1` was removed in cloudvision 1.21.0 and later. Might need to switch to `arista.tag.v2`.

KamilLelonek:
@glennmatthews yep, good catch - fixed in [`a2ccb18`](https://github.com/nautobot/nautobot-app-chatops/pull/445/changes/a2ccb18fcd32607c2e6fb08ed789aa963ae00383).

Bumped `cloudvision` to `>=1.25` and `protobuf` to `>=5` for Nautobot 3.x, and since that drops `arista.tag.v1` I moved `get_device_tags` over to `v2`. It's not just a rename - `DeviceTagAssignment*` becomes `TagAssignment*` with `element_type=ELEMENT_TYPE_DEVICE`, but `device_id`/`label`/`value` are the same so parsing didn't change.

I couldn't test against a real CloudVision, and I'm not 100% sure v2 needs `workspace_id` set to read active tags (left it unset like before).

Can you smoke-test the tag commands that'd confirm it?

glennmatthews (approved):
Seems plausible as far as the dependency updates go. Others will hopefully be able to help with the cloudvision usage and verification.

